PR #2, opened 2020-01-10
Ra dspdc 641 encode workflow

raaidbroad:
Write core logic of encode workflow for s3 to gcs, still needs to handle storage size correctly (it is hard coded, it shouldn't be). Not sure if I should add the role.yaml tbh

### DIFF
--- a/src/argo/workflow.yaml
+++ b/src/argo/workflow.yaml
@@ -22,6 +22,7 @@ spec:
         resources:
           requests:
             # NOTE this needs to be programmatically inserted, not hard coded; we are unable to do so at time of writing
+            # follow here: https://github.com/argoproj/argo/issues/1932
             storage: 100Gi
   templates:
     ## "Main" function for the extraction workflow.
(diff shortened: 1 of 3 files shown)


src/argo/workflow.yaml
```diff
@@ -0,0 +1,145 @@
...
+        secretName: "{{workflow.parameters.aws}}"
+  volumeClaimTemplates:
+    - metadata:
+        name: state-dir
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            # NOTE this needs to be programmatically inserted, not hard coded; we are unable to do so at time of writing
```
danxmoran: Can you add the URL to the issue?

src/argo/workflow.yaml
```diff
@@ -0,0 +1,145 @@
...
+          - name: aws-secret-volume
+          - name: aws-id-key
+          - name: aws-secret-key
+          - name: local-path
+      container:
+        image: us.gcr.io/broad-dsp-gcr-public/s3cmd-lite:latest
+        command: [s3cmd, sync]
+        args:
+          - s3://{{inputs.parameters.s3-bucket}}/{{inputs.parameters.s3-path}}
+          # NOTE that having slashes at the end of these is apparently consequential for a directory or file download
```
danxmoran: Can you add a link to the source docs that say this?

PR #3, opened 2020-01-23
Ra dspdc 642 start new pipeline

raaidbroad:
Start Encode pipeline, have it hit the biosamples endpoint and then pull the relevant donor and library information. Note that a bunch of this was copied and pasted from monster-etl's encode project and then updated (or that this code's structure was based on the code in that project).

### DIFF
--- a/src/main/scala/org/broadinstitute/monster/encode/extraction/EncodeEntity.scala
+++ b/src/main/scala/org/broadinstitute/monster/encode/extraction/EncodeEntity.scala
@@ -2,213 +2,28 @@ package org.broadinstitute.monster.encode.extraction
 
 import enumeratum.{Enum, EnumEntry}
 
+import scala.collection.immutable
+
 /**
   * An entity type found in ENCODE that we also want
   * (with modifications) in our data model.
   */
-sealed trait EncodeEntity extends EnumEntry {
-
-  /**
-    * Fields from the original ENCODE JSON objects of the
-    * entity type which should be retained.
-    *
-    * Assumption: trimming happens before renaming in the ETL workflow,
-    * so these strings should match the ENCODE names.
-    */
-  def fieldsToKeep: Set[String] = Set(
-    "@id",
-    "accession",
-    "aliases",
-    "award",
-    "date_created",
-    "dbxrefs",
-    "lab",
-    "status",
-    "submitted_by"
-  )
-
-  /**
-    * `old` -> `new` mappings for fields from ENCODE which
-    * should be renamed.
-    */
-  def fieldsToRename: List[(String, String)] = List(
-    "@id" -> "label",
-    "accession" -> "close_match",
-    "award" -> "sponsor",
-    "date_created" -> "created_at"
-  )
-
-  /**
-    * ENCODE fields which should be turned into links back to the ENCODE site.
-    *
-    * Assumption: linking happens after renaming, so these strings should match our
-    * data model's names.
-    */
-  def linkFields: Set[String] = Set("close_match", "sponsor", "submitted_by")
-
-  /**
-    * ENCODE fields of the form '/<entity-type>/<entity-label>/' which should be
-    * transformed into just '<entity-label>'.
-    *
-    * Assumption: labeling happens after renaming, so these strings should match our
-    * data model's names.
-    */
-  def labelFields: Set[String] = Set("label", "lab")
-
-  /**
-    * Fields which should be collapsed into a single "aliases" array.
-    *
-    * Assumption: collapsing happens after renaming, so these strings should match
-    * our data model's names.
-    */
-  def aliasFields: Set[String] = Set("aliases", "dbxrefs")
-
-  /**
-    * String ID for the entity type that will be queried
-    */
-  def encodeApiName: String
-}
+sealed trait EncodeEntity extends EnumEntry
 
 object EncodeEntity extends Enum[EncodeEntity] {
-  override val values = findValues
+  case object Biosample extends EncodeEntity
 
-  case object Biosample extends EncodeEntity {
+  case object Donor extends EncodeEntity
 
-    override def fieldsToKeep: Set[String] = super.fieldsToKeep.union(
-      Set(
-        "biosample_ontology",
-        "cell_isolation_method",
-        "date_obtained",
-        "source"
-      )
-    )
+  case object Experiment extends EncodeEntity
 
-    override def fieldsToRename: List[(String, String)] =
-      ("biosample_ontology" -> "biosample_type_id") :: super.fieldsToRename
+  case object File extends EncodeEntity
 
-    override def labelFields: Set[String] = super.labelFields.union(
-      Set(
-        "biosample_type_id",
-        "source"
-      )
-    )
+  case object Library extends EncodeEntity
 
-    override def encodeApiName: String = "Biosample"
-  }
+  case object Replicate extends EncodeEntity
 
-  case object Donor extends EncodeEntity {
+  case object Audit extends EncodeEntity
 
-    override def fieldsToKeep: Set[String] = super.fieldsToKeep.union(
-      Set(
-        "age_units",
-        "age",
-        "ethnicity",
-        "external_ids",
-        "health_status",
-        "life_stage",
-        "organism",
-        "sex"
-      )
-    )
-
-    override def fieldsToRename: List[(String, String)] = super.fieldsToRename ::: List(
-      "health_status" -> "phenotype",
-      "organism" -> "organism_id"
-    )
-
-    override def aliasFields: Set[String] = super.aliasFields + "external_ids"
-
-    override def labelFields: Set[String] = super.labelFields + "organism_id"
-
-    override def encodeApiName: String = "HumanDonor"
-  }
-
-  case object Experiment extends EncodeEntity {
-
-    override def fieldsToKeep: Set[String] = super.fieldsToKeep.union(
-      Set(
-        "date_released",
-        "date_submitted",
-        "description",
-        "target"
-      )
-    )
-
-    override def linkFields: Set[String] = super.linkFields + "target"
-
-    override def encodeApiName = "Experiment"
-  }
-
-  case object File extends EncodeEntity {
-
-    override def fieldsToKeep: Set[String] = super.fieldsToKeep.union(
-      Set(
-        "file_format",
-        "file_format_type",
-        "file_size",
-        "md5sum",
-        "notes",
-        "output_type",
-        "paired_end",
-        "platform",
-        "read_count",
-        "read_length",
-        "run_type",
-        "s3_uri",
-        "status"
-      )
-    )
-
-    override def fieldsToRename: List[(String, String)] = super.fieldsToRename ::: List(
-      "file_format" -> "format",
-      "file_format_type" -> "file_sub_type",
-      "paired_end" -> "paired_end_identifier",
-      "platform" -> "platform_id",
-      "run_type" -> "paired_end",
-      "s3_uri" -> "file_path"
-    )
-
-    override def aliasFields: Set[String] = super.aliasFields + "external_ids"
-
-    override def labelFields: Set[String] = super.labelFields + "platform_id"
-
-    override def encodeApiName = "File"
-  }
-
-  case object Library extends EncodeEntity {
-    override def fieldsToKeep: Set[String] = super.fieldsToKeep + "strand_specificity"
-
-    override def fieldsToRename: List[(String, String)] =
-      ("strand_specificity" -> "strand_specific") :: super.fieldsToRename
-
-    override def encodeApiName = "Library"
-  }
-
-  case object Replicate extends EncodeEntity {
-    override def fieldsToKeep: Set[String] = ???
-
-    override def fieldsToRename: List[(String, String)] = ???
-
-    override def linkFields: Set[String] = ???
-
-    override def labelFields: Set[String] = ???
-
-    override def aliasFields: Set[String] = ???
-
-    override def encodeApiName = "Replicate"
-  }
-
-  case object Audit extends EncodeEntity {
-    override def fieldsToKeep: Set[String] = ???
-
-    override def fieldsToRename: List[(String, String)] = ???
-
-    override def linkFields: Set[String] = ???
-
-    override def labelFields: Set[String] = ???
-
-    override def aliasFields: Set[String] = ???
-
-    override def encodeApiName = "File"
-  }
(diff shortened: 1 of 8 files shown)


src/main/scala/org/broadinstitute/monster/etl/encode/EncodeEntity.scala
```diff
@@ -0,0 +1,214 @@
+package org.broadinstitute.monster.etl.encode
+
+import enumeratum.{Enum, EnumEntry}
+
+/**
+  * An entity type found in ENCODE that we also want
+  * (with modifications) in our data model.
+  */
+sealed trait EncodeEntity extends EnumEntry {
```
danxmoran: I'm not sure we want to repeat this pattern in the pipeline redo. Can we delete it for now? We definitely don't need it for the extraction portion.

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeExtractions.scala
```diff
@@ -0,0 +1,216 @@
...
+  /** HTTP client to use for querying ENCODE APIs. */
+  val client = new OkHttpClient()
+
+  /**
+    * Pipeline stage which maps batches of query params to output payloads
+    * by sending the query params to the ENCODE search API.
+    *
+    * @param encodeEntity the type of ENCODE entity the stage should query
+    */
+  class EncodeLookup(encodeEntity: EncodeEntity)
```
danxmoran: If `EncodeEntity` goes away, this will be a `String`

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeEntity.scala
danxmoran: Alt: We could keep an `EncodeEntity` but delete all the methods / hard-coded collections of strings here. Then we'll just have:
```scala
sealed trait EncodeEntity extends EnumEntry
object EncodeEntity extends Enum[EncodeEntity] {
  case object Biosample extends EncodeEntity
  // etc.
}
```

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeExtractions.scala
```diff
@@ -0,0 +1,216 @@
...
+
+    override def asyncLookup(
+      client: OkHttpClient,
+      params: List[(String, String)]
+    ): Future[String] = {
+      val paramStrings = params.map {
+        case (key, value) =>
+          s"$key=$value"
+      }
+      val allParams = s"type=${encodeEntity.encodeApiName}" :: baseParams ::: paramStrings
```
danxmoran: If `EncodeEntity` stays but all methods are deleted, this would be `encodeEntity.entryName`

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeExtractions.scala
```diff
@@ -0,0 +1,216 @@
...
+
+      get(
+        client,
+        allParams
+      )
```
danxmoran: I normally don't want to fight the formatter, but could we push this onto one line?

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeExtractions.scala
```diff
@@ -0,0 +1,216 @@
...
+        Option(value).foreach { jsonObj =>
+          val bytes = jsonObj.asJson.noSpaces.getBytes
+          VarInt.encode(bytes.length, outStream)
+          outStream.write(bytes)
+        }
+      override def decode(inStream: InputStream): JsonObject = {
+        val numBytes = VarInt.decodeInt(inStream)
+        val bytes = new Array[Byte](numBytes)
+        ByteStreams.readFully(inStream, bytes)
+        parser.decodeByteBuffer[JsonObject](ByteBuffer.wrap(bytes)).right.get
```
danxmoran: For now, instead of `.right.get` would we `.fold(throw _, identity)`?

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeExtractions.scala
```diff
@@ -0,0 +1,216 @@
...
+    */
+  def getEntitiesByField(
+    encodeEntity: EncodeEntity,
+    fieldName: String = "@id"
+  ): SCollection[String] => SCollection[JsonObject] = { idStream =>
+    val paramsBatchStream =
+      idStream.transform(s"Build ${encodeEntity.entryName} ID Queries") {
+        _.map(KV.of("key", _))
+          .setCoder(KvCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of()))
+          .applyKvTransform(GroupIntoBatches.ofSize(100))
```
danxmoran: Do you think it's worth making the batch size an input parameter?

src/main/scala/org/broadinstitute/monster/etl/encode/ExtractionPipeline.scala
```diff
@@ -0,0 +1,72 @@
+package org.broadinstitute.monster.etl.encode
```
danxmoran: Nitpick: could we use `org.broadinstitute.monster.encode.extraction` as the package? I've been using `org.broadinstitute.monster.clinvar` as the package in the ClinVar code migration.

src/main/scala/org/broadinstitute/monster/etl/encode/ExtractionPipeline.scala
```diff
@@ -0,0 +1,72 @@
...
+        referenceField = "accession",
+        manyReferences = false
+      )(biosamples)
+
+      EncodeExtractions.getEntitiesByField(EncodeEntity.Library, "biosample.accession")(
+        biosampleIdParams
+      )
+    }
+
+    biosamples.saveAsJsonFile(s"${parsedArgs.outputDir}/biosamples")
```
danxmoran: Can we use the raw entity name as the directory here? It's the same pattern we use for extracted XML.

src/main/scala/org/broadinstitute/monster/etl/encode/ExtractionPipeline.scala
```diff
@@ -0,0 +1,72 @@
...
+    */
+  def main(rawArgs: Array[String]): Unit = {
+    val (pipelineContext, parsedArgs) = ContextAndArgs.typed[Args](rawArgs)
+
+    val biosamples = {
+      val additionalParams =
+        pipelineContext.parallelize(List(List("organism.name" -> "human")))
+
+      EncodeExtractions.getEntities(EncodeEntity.Biosample)(additionalParams)
+    }
```
danxmoran: What do you think of this block pattern? I'm not sure I like it, but I could be convinced if others find it useful.

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeEntity.scala
raaidbroad: Do we not intend to have fields to keep or rename or anything? Or do we intend to implement that differently moving forward?

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeExtractions.scala
raaidbroad: Probably; input parameter from the command line, right?

src/main/scala/org/broadinstitute/monster/etl/encode/ExtractionPipeline.scala
raaidbroad: I think it's nice and tidy? Would the alternative be to just not have them in blocks and like assign the output of line 40 to `val biosamples`?

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeEntity.scala
danxmoran: I'm guessing we'll implement it more like the ClinVar pipeline.

src/main/scala/org/broadinstitute/monster/etl/encode/EncodeExtractions.scala
danxmoran: Yes, another command-line arg. You should be able to set a default of 100 to keep the default behavior the same

src/main/scala/org/broadinstitute/monster/etl/encode/ExtractionPipeline.scala
danxmoran: We could use the curry-ing built into the `getEntities` signature to inline things, like this:
```scala
val biosamples = EncodeExtractions.getEntities(EncodeEntity.Biosample) {
  pipelineContext.parallelize(List(List("organism.name" -> "human")))
}
```
I'm also wondering if there's enough repetition here to make a generic method that we call over and over again. I imagine that method would also include the `saveAsJsonFile` piece.

src/main/scala/org/broadinstitute/monster/encode/extraction/EncodeEntity.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  case object File extends EncodeEntity
+
+  case object Library extends EncodeEntity
+
+  case object Replicate extends EncodeEntity
+
+  case object Audit extends EncodeEntity
+
+  override def values: immutable.IndexedSeq[EncodeEntity] = ???
```
danxmoran:
```suggestion
  override def values: immutable.IndexedSeq[EncodeEntity] = findValues
```

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -0,0 +1,112 @@
...
+     * Generic helper method for extracting linked entities and saving them.
+     *
+     * @param encodeEntity the entity from which to extract a field
+     * @param idEntity the entity whose IDs to use as input to the getEntitiesByField
+     * @param idReferenceField the field name to use as an identifier in getIds
+     * @param prevData the data to base the ID extraction off of
+     * @param entityFieldName the field name to extract if it isn't @id
+     * @return the extracted linked entities
+     */
+    def extractLinkedEntity(
```
danxmoran:
```suggestion
    def extractLinkedEntities(
```

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -0,0 +1,112 @@
...
+    ): SCollection[JsonObject] = {
+      val out = EncodeExtractions.getEntitiesByField(
+        encodeEntity,
+        parsedArgs.batchSize,
+        entityFieldName
+      ) {
+        EncodeExtractions.getIds(
+          idEntity.entryName,
+          idReferenceField,
+          false
```
danxmoran: I don't think this will always be hard-code-able to `false`

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -0,0 +1,112 @@
...
+     * @param encodeEntity the entity from which to extract a field
+     * @param idEntity the entity whose IDs to use as input to the getEntitiesByField
+     * @param idReferenceField the field name to use as an identifier in getIds
+     * @param prevData the data to base the ID extraction off of
+     * @param entityFieldName the field name to extract if it isn't @id
+     * @return the extracted linked entities
+     */
+    def extractLinkedEntity(
+      encodeEntity: EncodeEntity,
+      idEntity: EncodeEntity,
```
danxmoran: The names of these are confusing to me. Do we really need both here? It looks like `idEntity` is only used in a `transform` call which we could factor away.

src/main/scala/org/broadinstitute/monster/encode/extraction/EncodeExtractions.scala
```diff
@@ -0,0 +1,217 @@
...
+    * @param referenceField field in the input JSONs containing the IDs
+    *                       to extract
+    * @param manyReferences whether or not `referenceField` is an array
+    */
+  def getIds(
+    entryName: String,
+    referenceField: String,
+    manyReferences: Boolean
+  ): SCollection[JsonObject] => SCollection[String] =
+    _.transform(s"Get $entryName IDs") { collection =>
```
danxmoran: I would ditch this `transform` call in favor of more specific calls at the top of the pipeline.

src/main/scala/org/broadinstitute/monster/encode/extraction/EncodeExtractions.scala
```diff
@@ -0,0 +1,217 @@
...
+  /**
+    * Pipeline stage which maps batches of search parameters into JSON entities
+    * from ENCODE matching those parameters.
+    *
+    * @param encodeEntity the type of ENCODE entity the stage should query
+    */
+  def getEntities(
+    encodeEntity: EncodeEntity
+  ): SCollection[List[(String, String)]] => SCollection[JsonObject] =
+    _.transform(s"Download ${encodeEntity.entryName} Entities") {
```
danxmoran: Would ditch this `transform`

src/main/scala/org/broadinstitute/monster/encode/extraction/EncodeExtractions.scala
```diff
@@ -0,0 +1,217 @@
...
+    * @param batchSize the number of elements in a batch stream
+    * @param fieldName the field name to get the entity by, is "@id" by default
+    */
+  def getEntitiesByField(
+    encodeEntity: EncodeEntity,
+    batchSize: Long,
+    fieldName: String = "@id"
+  ): SCollection[String] => SCollection[JsonObject] = { idStream =>
+    val paramsBatchStream =
+      idStream.transform(s"Build ${encodeEntity.entryName} ID Queries") {
```
danxmoran: Would ditch this `transform`

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -0,0 +1,112 @@
...
+
+    val replicates = extractLinkedEntity(
+      EncodeEntity.Replicate,
+      EncodeEntity.Library,
+      "accession",
+      libraries,
+      "library.accession"
+    )
+
+    // don't need to use experiments apart from storing them, so we don't assign an output here
```
danxmoran: There is one more ticket in the backlog to chain experiments -> files, so I would replace this with a TODO

PR #5, opened 2020-02-03
DSPDC-725 Initial Commit

quazi-broad:
Adding a new object to the EncodeEntity collection to add FunctionalCharacterizationExperiment. This dataset was previously assumed to be included under the "type=Experiment" query.
Also added code to the ExtractionPipeline script to store a new parent entity FunctionalCharacterizationExperiment and also added code to ensure we're pulling file data from that set of experiments.

### DIFF
--- a/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
+++ b/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
@@ -118,7 +118,7 @@ object ExtractionPipeline {
     extractLinkedEntities(
       entityToExtract = EncodeEntity.File,
       matchingField = "@id",
-      linkingEntities = SCollection.unionAll([experiments, fcExperiments]),
+      linkingEntities = SCollection.unionAll(List(experiments, fcExperiments)),
       linkedField = "dataset"
     )
 
--- a/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
+++ b/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
@@ -118,7 +118,7 @@ object ExtractionPipeline {
     extractLinkedEntities(
       entityToExtract = EncodeEntity.File,
       matchingField = "@id",
-      linkingEntities = SCollection.unionAll(experiments, fcExperiments),
+      linkingEntities = SCollection.unionAll([experiments, fcExperiments]),
       linkedField = "dataset"
     )
 
--- a/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
+++ b/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
@@ -108,7 +108,7 @@ object ExtractionPipeline {
       linkingEntities = replicates
     )
 
-    val fc_experiments = extractLinkedEntities(
+    val fcExperiments = extractLinkedEntities(
       entityToExtract = EncodeEntity.FunctionalCharacterizationExperiment,
       matchingField = "experiment",
       linkingEntities = replicates
@@ -118,15 +118,7 @@ object ExtractionPipeline {
     extractLinkedEntities(
       entityToExtract = EncodeEntity.File,
       matchingField = "@id",
-      linkingEntities = experiments,
-      linkedField = "dataset"
-    )
-
-    // don't need to use files apart from storing them, so we don't assign an output here
-    extractLinkedEntities(
-      entityToExtract = EncodeEntity.File,
-      matchingField = "@id",
-      linkingEntities = fc_experiments,
+      linkingEntities = SCollection.unionAll(experiments, fcExperiments),
       linkedField = "dataset"
     )
 
--- a/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
+++ b/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
@@ -102,16 +102,22 @@ object ExtractionPipeline {
       linkedField = "library.accession"
     )
 
+    //[partition] that stream to separate [type=experiment]
+    // from [type=functional-characterization-experiment]
+    val (expReplicate, fcReplicate) = replicates.partition {
+      replicate => replicate("experiment").exists(experimentString => experimentString.equals("Experiment"))
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -116,6 +122,14 @@ object ExtractionPipeline {
       linkedField = "dataset"
     )
 
+    // don't need to use files apart from storing them, so we don't assign an output here
+    extractLinkedEntities(
+      entityToExtract = EncodeEntity.File,
+      matchingField = "@id",
+      linkingEntities = fc_experiments,
+      linkedField = "dataset"
+    )
```
danxmoran: I can't tell for sure from the code that's in-view, but I don't think this is safe as-written. Both this call and the existing one above will try to write their outputs under the same prefix. I think it's possible they'll step on each other.

I would `SCollection.unionAll(experiments, fc_experiments)` to get the combination of both streams, and feed that into this `extractLinkedEntities`

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -108,6 +108,12 @@ object ExtractionPipeline {
       linkingEntities = replicates
     )
 
+    val fc_experiments = extractLinkedEntities(
```
danxmoran:
```suggestion
    val fcExperiments = extractLinkedEntities(
```
Scala style

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
quazi-broad: Okay, that makes sense - I was wondering if I could just list both experiments and fcExperiments but was worried that wouldn't work since extractLinkedEntities is expecting a SCollection[JsonObject] for the linkingEntities parameter. Thank you!

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
quazi-broad: Switched this to "fcExperiments" throughout the script

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -108,11 +108,17 @@ object ExtractionPipeline {
...
+      matchingField = "experiment",
+      linkingEntities = replicates
+    )
+
     // don't need to use files apart from storing them, so we don't assign an output here
     extractLinkedEntities(
       entityToExtract = EncodeEntity.File,
       matchingField = "@id",
-      linkingEntities = experiments,
+      linkingEntities = SCollection.unionAll([experiments, fcExperiments]),
```
danxmoran: Looks like you've got Python on the brain!

In Scala you build collections via `apply` on their companion objects. For example, this could be:
```suggestion
      linkingEntities = SCollection.unionAll(List(experiments, fcExperiments)),
```

quazi-broad: Thanks! Just figured that out with Raaid the hard way lol.

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -108,11 +108,17 @@ object ExtractionPipeline {
       linkingEntities = replicates
     )
 
+    val fcExperiments = extractLinkedEntities(
+      entityToExtract = EncodeEntity.FunctionalCharacterizationExperiment,
+      matchingField = "experiment",
+      linkingEntities = replicates
```
danxmoran: This should have the correct behavior as-is, but it's sub-optimal. There will be batches of IDs that contain no `/experiment/`s, and other batches that contain no `/functional-characterization-experiment/`s. Those queries are guaranteed to return 0 results when we run them against the wrong `type=` value, so it'd be good for us (in runtime) and ENCODE (in server stability) if we avoided making them.

`SCollection` has a few `partition` methods that I think would be useful here: https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L262-L286. I think to reuse the plumbing, the flow will need to be:
1. `flatMap` the `replicates` stream so we end up with an `SCollection` of `JsonObject("experiment" -> "/<type>/<accession>")`
2. `partition` that stream to separate `type=experiment` from `type=functional-characterization-experiment`
3. Feed each partition into `extractLinkedEntities` calls for the corresponding type

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -102,17 +102,29 @@ object ExtractionPipeline {
       linkedField = "library.accession"
     )
 
+    //[partition] that stream to separate [type=experiment]
+    // from [type=functional-characterization-experiment]
+    val (expReplicate, fcReplicate) = replicates.partition {
+      replicate => replicate("experiment").exists(experimentString => experimentString.equals("Experiment"))
```
danxmoran: I think this is off. `experimentString` has type `Json`, so `equals` will always return `false` when the RHS is a `String`. The values of the `"experiment"` are also going to be IDs, so I don't think we can check against a fixed constant.

I'm reconsidering the anonymous function usage. Since this is just an optimization it's tough to see if this logic is working based on pipeline outputs, and it's tough to unit test when it's embedded inside the `partition` call. I'd split out a new method:
```scala
def isFunctionalCharacterizationReplicate(replicate: JsonObject): Boolean = ???
```
We can unit-test the method using some cherry-picked examples pulled from ENCODE.

quazi-broad: Added a unit test!

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
quazi-broad: Created a new method to determine whether something is a functional-characterization-experiment or not. The method uses flatMap to flatten the replicate stream and turn it into a string and then evaluates the contents of the string. This method is working properly and correctly partitioning the replicates stream.

PR #6, opened 2020-02-06
Enable tests, with coverage, in CI.

danxmoran:
This might fail since we don't have any tests / integration-tests right now... let's see what happens!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,7 +34,7 @@ pipeline {
     }
     post {
         always {
-            junit '**/target/test-reports/*'
+            junit allowEmptyResults: true, testResults: '**/target/test-reports/*'
         }
         success {
             script {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -21,21 +21,44 @@ pipeline {
...
+                sh 'sbt "set ThisBuild/coverageEnabled := true" IntegrationTest/test'
+            }
+        }
     }
     post {
-        //always {
-        //    junit '**/target/test-reports/*'
-        //}
+        always {
+            junit allowEmptyResults: true, testResults: '**/target/test-reports/*'
```
quazi-broad: Is the purpose of setting allowEmptyResults so that we can create unit tests without an output or maybe create placeholder unit tests? I'm wondering if we'd ever have an issue where an error within the unit test would result in an empty result.

danxmoran: It's to handle the current case where there are no tests defined / if the format-check or compile stages fail and tests never run. Without "allowEmptyResults" another error will show up in the logs, which can be a red herring to the actual problem.

PR #7, opened 2020-02-13
Extend encode extraction to include targets

snwessel:
Added Targets to the encode extraction and tried to improve some comments for the `extractLinkedEntities` function. 

The original [ticket description](https://broadinstitute.atlassian.net/browse/DSPDC-766) had suggested that we pull Antibody entities from the Replicates, and then pull Targets from the Antibodies. Raaid and I talked about this offline and weren't able to find anything that links Antibodies to Replicates (unless anyone knows otherwise) so we decided to change our approach. 

The new plan is to pull Targets from Experiments, and then pull Antibodies from Targets (the second part will be done in another ticket).

### DIFF
--- a/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
+++ b/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
@@ -43,11 +43,11 @@ object ExtractionPipeline {
       * Generic helper method for extracting linked entities and saving them.
       *
       * @param entityToExtract the entity which should be extracted
-      * @param matchingField name of the field to use as an identifier for the entityToExtract
-      *                      (should match the id that's used in the linkedField)
+      * @param matchingField name of the linkingEntity field which matches the value of the given "linkedField"
       * @param linkingEntities the parent entity (which references a set of encodeEntities)
       *                        from which ID extraction will be based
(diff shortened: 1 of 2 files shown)


snwessel:
Actually, I just found an example of a replicate with an antibody! So we can stick with the original plan after-all (sorry for the confusion!). I am going to close this PR for now and open it back up once we have merged changes to extract the antibodies.

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -116,6 +116,13 @@ object ExtractionPipeline {
       linkedField = "dataset"
     )
 
+    // don't need to use targets apart from storing them, so we don't assign an output here
+    extractLinkedEntities(
+      entityToExtract = EncodeEntity.Target,
+      matchingField = "target",
+      linkingEntities = experiments
```
quazi-broad: Just wondering - when "linkedField" is not provided, it will default to using the @id, correct?
A "type=Target" query has the following in the id field:
**"@id": "/targets/ztf-8-celegans/"**

snwessel: Yeah, that sounds right. And the target "@id" field should match up with the experiment "target" field.

PR #8, opened 2020-02-13
Add AntibodyLot entities to encode extraction

snwessel:
Adding AntibodyLots to the encode extraction ([Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-768)).

### DIFF
--- a/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
+++ b/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
@@ -108,12 +108,19 @@ object ExtractionPipeline {
       linkingEntities = replicates
     )
 
-    // don't need to use files apart from storing them, so we don't assign an output here
+    // don't need to use the following entities apart from storing them, so we don't assign an output here
+
     extractLinkedEntities(
       entityToExtract = EncodeEntity.File,
       matchingField = "@id",
       linkingEntities = experiments,
       linkedField = "dataset"
+    )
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -128,14 +128,21 @@ object ExtractionPipeline {
       linkingEntities = fcReplicate
     )
 
-    // don't need to use files apart from storing them, so we don't assign an output here
+    // don't need to use the following entities apart from storing them, so we don't assign an output here
```
danxmoran: IMO we should just remove this comment, because we keep having to update it as the workflow evolves

src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipeline.scala
```diff
@@ -128,14 +128,21 @@ object ExtractionPipeline {
...
       matchingField = "@id",
       linkingEntities = SCollection.unionAll(List(experiments, fcExperiments)),
       linkedField = "dataset"
     )
 
+    extractLinkedEntities(
+      entityToExtract = EncodeEntity.AntibodyLot,
+      matchingField = "antibody",
+      linkingEntities = replicates
+    )
```
danxmoran: Can you move this up to be right under the `replicates` definition, before the partitioning block?

PR #10, opened 2020-02-20
QH DSPDC-765 - Update encode-ingest to use new scio-utils library

quazi-broad:
Added the sbt-plugins-scio plugin + bumped up the Jade plugin to match (0.7.0).
Replaced MonsterJadeDatasetPlugin with MonsterScioPipelinePlugin.
Updated EncodeExtractions and ExtractionPipeline to replace usages of JsonObject (io.circe) with Msg (upack). This required some rework - using the UpackMsgCoder, replacing the SCollection[JsonObject] with SCollection[Msg], and replacing `extract` calls with `read` or `tryRead` and instance of `saveAsJsonFile` with `writeJsonLists`.
For the unit tests, I used a JsonParser to read the mocked JsonObjects and pass them to the tested function.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,6 @@ lazy val `encode-ingest` = project
     ),
     libraryDependencies ++= Seq(
       "org.scalatest" %% "scalatest" % scalatestVersion,
-      "io.circe" %% "circe-literal" % MonsterJadeDatasetPlugin.CirceVersion,
     ).map(_ % Test),
     scioReleaseBucketName := "TODO",
     scioSnapshotBucketName := "TODO"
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -1,28 +1,18 @@
...
-    jadeTablePackage := "org.broadinstitute.monster.encode.jadeschema.table",
-    jadeStructPackage := "org.broadinstitute.monster.encode.jadeschema.struct",
     libraryDependencies ++= Seq(
       "com.beachape" %% "enumeratum" % enumeratumVersion,
-      "ch.qos.logback" % "logback-classic" % logbackVersion,
-      "com.spotify" %% "scio-extra" % scioVersion,
     ),
     libraryDependencies ++= Seq(
       "org.scalatest" %% "scalatest" % scalatestVersion,
       "io.circe" %% "circe-literal" % MonsterJadeDatasetPlugin.CirceVersion,
```
danxmoran: You should be able to remove this dependency now that the test isn't using it.

PR #15, opened 2020-03-03
split project into multi-project build

raaidbroad:
What the title says


build.sbt
```diff
@@ -1,13 +1,30 @@
-val enumeratumVersion = "1.5.13"
+import org.broadinstitute.monster.sbt.model.JadeIdentifier
 
+val enumeratumVersion = "1.5.13"
 val okhttpVersion = "4.4.0"
 
 lazy val `encode-ingest` = project
   .in(file("."))
+  .aggregate(`encode-extraction`, `encode-transformation`)
```
danxmoran: I think you also want a line for:
```scala
.settings(publish / skip := true)
```

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
```diff
@@ -1,10 +1,10 @@
 package org.broadinstitute.monster.encode.extraction
 
 import caseapp.{AppName, AppVersion, HelpMessage, ProgName}
-import org.broadinstitute.monster.EncodeIngestBuildInfo
```
quazi-broad: I like that you renamed this to be extraction specific since "Ingest" could be confusing.
We're only adding transformation and orchestration pieces, right?

raaidbroad: yeah I'm following the format Dan made for the dog-aging-ingest, which has extraction and transformation. One of our plugins generates the BuildInfo thing so it updates the name based on the project 🙂 very useful

PR #17, opened 2020-03-10
QH - DSPDC-763 - Parameterized the ENCODE extraction workflow + Added integration test

quazi-broad:
The Encode extraction workflow can now take in query parameters.
This will allow us to pass more targeted entry-points to the pipeline during testing and also allow us to query for other organism types (currently hardcoded to pull human donor data).
I also added an integration test that just takes one biosample accession and makes sure that we downloaded some stuff to the temporary directories. Not sure how often biosamples change - it would be great to use a biosample we know will never change. Maybe Encode has some seed data for this purpose?

### DIFF
--- a/extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineIntegrationSpec.scala
+++ b/extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineIntegrationSpec.scala
@@ -24,61 +24,56 @@ class ExtractionPipelineIntegrationSpec extends PipelineBuilderSpec[Args] {
   override val builder = ExtractionPipeline.pipelineBuilder
 
   // Paths for entity files downloaded
-  val AnalysisStepPath = (outputDir.pathAsString + "/AnalysisStep")
-  val AnalysisStepRunPath = (outputDir.pathAsString + "/AnalysisStepRun")
-  val AnalysisStepVersionPath = (outputDir.pathAsString + "/AnalysisStepVersion")
-  val AntibodyLotPath = (outputDir.pathAsString + "/AntibodyLot")
-  val BiosamplePath = (outputDir.pathAsString + "/Biosample")
-  val DonorPath = (outputDir.pathAsString + "/Donor")
-  val ExperimentPath = (outputDir.pathAsString + "/Experiment")
-  val FilePath = (outputDir.pathAsString + "/File")
-  val FunctionalCharacterizationExperimentPath = (outputDir.pathAsString + "/FunctionalCharacterizationExperiment")
-  val LibraryPath = (outputDir.pathAsString + "/Library")
-  val ReplicatePath = (outputDir.pathAsString + "/Replicate")
-  val TargetPath = (outputDir.pathAsString + "/Target")
+  val analysisStepPath = (outputDir.pathAsString + "/AnalysisStep")
+  val analysisStepRunPath = (outputDir.pathAsString + "/AnalysisStepRun")
+  val analysisStepVersionPath = (outputDir.pathAsString + "/AnalysisStepVersion")
+  val antibodyLotPath = (outputDir.pathAsString + "/AntibodyLot")
+  val biosamplePath = (outputDir.pathAsString + "/Biosample")
+  val donorPath = (outputDir.pathAsString + "/Donor")
+  val experimentPath = (outputDir.pathAsString + "/Experiment")
+  val filePath = (outputDir.pathAsString + "/File")
+  val functionalCharacterizationExperimentPath = (outputDir.pathAsString + "/FunctionalCharacterizationExperiment")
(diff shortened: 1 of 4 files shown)


extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineIntegrationSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+  val AnalysisStepVersionPath = (outputDir.pathAsString + "/AnalysisStepVersion")
+  val AntibodyLotPath = (outputDir.pathAsString + "/AntibodyLot")
+  val BiosamplePath = (outputDir.pathAsString + "/Biosample")
+  val DonorPath = (outputDir.pathAsString + "/Donor")
+  val ExperimentPath = (outputDir.pathAsString + "/Experiment")
+  val FilePath = (outputDir.pathAsString + "/File")
+  val FunctionalCharacterizationExperimentPath = (outputDir.pathAsString + "/FunctionalCharacterizationExperiment")
+  val LibraryPath = (outputDir.pathAsString + "/Library")
+  val ReplicatePath = (outputDir.pathAsString + "/Replicate")
+  val TargetPath = (outputDir.pathAsString + "/Target")
```
danxmoran: Style nitpick: it'd be more conventional for these to all be lower-camel-case, i.e.
```suggestion
  val targetPath = (outputDir.pathAsString + "/Target")
```
Upper-camel-case is typically reserved for static constants

extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineIntegrationSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  /// VERIFY DOWNLOADS
+  behavior of "ExtractionPipeline"
+
+  it should "successfully download results from the from Encode API" in {
+    outputDir.nonEmpty should be(true)
+    println("outputDir has contents...")
```
danxmoran: I'd drop all the `println`s, the normal test output should be sufficient

extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineIntegrationSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+  val FunctionalCharacterizationExperimentPath = (outputDir.pathAsString + "/FunctionalCharacterizationExperiment")
+  val LibraryPath = (outputDir.pathAsString + "/Library")
+  val ReplicatePath = (outputDir.pathAsString + "/Replicate")
+  val TargetPath = (outputDir.pathAsString + "/Target")
+
+  /// VERIFY DOWNLOADS
+  behavior of "ExtractionPipeline"
+
+  it should "successfully download results from the from Encode API" in {
+    outputDir.nonEmpty should be(true)
```
danxmoran: Does this check add anything over the more specific checks below?

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
```diff
@@ -20,5 +22,22 @@ case class Args(
...
+  batchSize: Long,
+  @HelpMessage(
+    "Initial query to target a specific entry-point to the pipeline."
+  )
+  initialQuery: List[(String, String)] = List("organism.name" -> "human")
 )
+
+object Args {
+
+  implicit val customArgParser: ArgParser[(String, String)] =
```
danxmoran: Now that it's working, would be more specific with the name
```suggestion
  implicit val tupleParser: ArgParser[(String, String)] =
```

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -67,7 +67,7 @@ class ExtractionPipelineBuilder(getClient: () => EncodeClient)
 
     // biosamples are the first one and follow a different pattern, so we don't use the generic method
     val biosamples = ctx
-      .parallelize(List(List("organism.name" -> "human")))
+      .parallelize(List({ args.initialQuery }))
```
danxmoran: Brackets not needed here
```suggestion
      .parallelize(List(args.initialQuery))
```

extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineIntegrationSpec.scala
quazi-broad: No, I actually had written this first to make sure the whole test script was working before I decided to go down a level and check within the individual entity files. I'll remove this check.

PR #18, opened 2020-03-10
Ra dspdc 237 extraction unit tests

raaidbroad:
I left the structure of the fakeIds thing with mapping thereafter intact in case we would want to generate more than one entry for the params, but if we don't intend to do that then I can remove that and clean up the code accordingly.


extraction/src/test/scala/org/broadinstitute/monster/encode/extraction/EncodeExtractionPipelineBuilderSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+
+import better.files.File
+import org.broadinstitute.monster.common.PipelineBuilderSpec
+import org.broadinstitute.monster.common.msg.MsgOps
+import upack._
+
+import scala.collection.mutable
+
+object EncodeExtractionPipelineBuilderSpec {
+  val fakeIds = 1 to 1
```
snwessel: Is the idea that this range could be made wider in the future? I am trying to understand the benefit of using a range when it's only `1 to 1`.

raaidbroad: Yeah lol that's the only reason I kept it as is; if we think that's not worth it then I can certainly change it.

danxmoran: I'd get rid of these for all the objects with only one result, for clarity

extraction/src/test/scala/org/broadinstitute/monster/encode/extraction/EncodeExtractionPipelineBuilderSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+          Str("accession") -> Str(i.toString)
+        ): Msg
+      }.to[mutable.ArrayBuffer]
+    )
+  )
+
+  // replicates
+  val replicateParams = ("library.accession", "1")
+
+  val replicateIds = 1 to 2
```
danxmoran: Should there be a downstream step that includes multiple expected queries because of this, or are we not chaining any objects off of replicates?

danxmoran: If nothing's being chained from replicates, can we bump one of the other objects to return multiple things, to be sure that chaining works properly?

extraction/src/test/scala/org/broadinstitute/monster/encode/extraction/EncodeExtractionPipelineBuilderSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+      (EncodeEntity.AntibodyLot, antibodyParams),
+      (EncodeEntity.Target, targetParams),
+      (EncodeEntity.Experiment, experimentParams),
+      (EncodeEntity.FunctionalCharacterizationExperiment, fcExperimentParams),
+      (EncodeEntity.File, fileParams),
+      (EncodeEntity.AnalysisStepRun, analysisStepRunParams),
+      (EncodeEntity.AnalysisStepVersion, analysisStepVersionParams),
+      (EncodeEntity.AnalysisStep, analysisStepParams)
+    )
+  }
```
danxmoran: Can we structure these checks to be more fine-grained, with one `it should` per entity type? We can bundle the `recordedRequests` and `readMsgs` checks into the same cases

extraction/src/test/scala/org/broadinstitute/monster/encode/extraction/EncodeExtractionPipelineBuilderSpec.scala
raaidbroad: experiments are chained from replicates; I have 2 here so that we get at least one thing for `experiment` and one for `fcexperiment

PR #19, opened 2020-03-12
Sw dspdc 310 initialize encode transformation

snwessel:
Initializing the transformation pipeline in Encode to map HumanDonor objects([Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-310)). 

_Also, I'm not sure why GitHub is showing 32 commits (most of the commit messages it's showing are a part of the json schema ticket, which has already been merged)._


transformation/src/jade-schema/tables/human_donor.table.json
```diff
@@ -48,8 +47,7 @@
     },
     {
       "name": "life_stage",
-      "datatype": "string",
-      "type": "required"
+      "datatype": "string"
     },
     {
       "name": "parent_ids",
```
snwessel: I realized that "age" and "life_stage" are not always defined in the encode data, so I am making them optional here. I ran this change by Kathy already.

transformation/src/jade-schema/tables/human_donor.table.json
```diff
@@ -18,8 +18,7 @@
     },
     {
       "name": "age",
-      "datatype": "string",
-      "type": "required"
+      "datatype": "string"
```
danxmoran: Should age be a string? Or an integer?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/Args.scala
```diff
@@ -0,0 +1,14 @@
...
+
+@AppName("Encode transformation pipeline")
+@AppVersion(EncodeTransformationBuildInfo.version)
+@ProgName("org.broadinstitute.monster.etl.encode.EncodeTransformationPipeline")
+case class Args(
+  @HelpMessage("Path to the top-level directory where JSON was extracted")
+  inputPrefix: String,
+  @HelpMessage("Path where transformed JSON should be written")
+  outputPrefix: String
+)
```
snwessel: For the most part, this file was copied over from ClinVar. Maybe it would make sense to move it into a common utility at some point?

transformation/src/jade-schema/tables/human_donor.table.json
danxmoran: (It might need to become 2 columns: age and age_unit)

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -0,0 +1,59 @@
...
+    ()
+  }
+
+  def transformDonor(donorInput: Msg): HumanDonor = {
+    HumanDonor(
+      id = donorInput.read[String]("accession"),
+      crossReferences = donorInput.read[Array[String]]("dbxrefs"),
+      timeCreated = donorInput.read[OffsetDateTime]("date_created"),
+      age = donorInput.tryRead[String]("age"),
+      ethnicities = Array() ++ donorInput.tryRead[String]("ethnicity"),
```
snwessel: I am using this messy syntax to convert an `Option[String]` to an `Array[String]`. Let me know if anyone knows of a better way to do this.

danxmoran: You can do this instead:
```suggestion
      ethnicities = donorInput.tryRead[Array[String]]("ethnicity").getOrElse(Array.empty),
```
Easier to understand the intent instead of the type-fu (at least to me)

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -0,0 +1,59 @@
...
+    )
+    ()
+  }
+
+  def transformDonor(donorInput: Msg): HumanDonor = {
+    HumanDonor(
+      id = donorInput.read[String]("accession"),
+      crossReferences = donorInput.read[Array[String]]("dbxrefs"),
+      timeCreated = donorInput.read[OffsetDateTime]("date_created"),
+      age = donorInput.tryRead[String]("age"),
```
danxmoran: If `age` gets split, this will need some more advanced parsing (fun!)

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -0,0 +1,59 @@
...
+      ethnicities = Array() ++ donorInput.tryRead[String]("ethnicity"),
+      organism = donorInput.read[String]("organism"),
+      sex = donorInput.read[String]("sex"),
+      award = donorInput.read[String]("award"),
+      lab = donorInput.read[String]("lab"),
+      lifeStage = donorInput.tryRead[String]("life_stage"),
+      parentIds = donorInput.read[Array[String]]("parents"),
+      status = donorInput.read[String]("status"),
+      twinIds = Array() ++ donorInput.tryRead[String]("twin"),
```
danxmoran: Same idea as ethnicities

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: Wait... if there's only ever one ethnicity, why does the column need to be an array? (Array vs. not makes no difference for search in elasticsearch, but it does make SQL-ing more complicated in BigQuery)

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: And same question about "why an array?"

transformation/src/test/test-files/outputs/human_donor/part1.json
```diff
@@ -0,0 +1,2 @@
+{"id":"ENCDO845WKR","cross_references":["GEO:SAMN05897796"],"time_created":"2015-02-14T00:31:58.410757Z","age":"37","ethnicities":[],"organism":"/organisms/human/","sex":"male","award":"/awards/ENCODE/","lab":"/labs/encode-consortium/","life_stage":"adult","parent_ids":[],"status":"released","twin_ids":[],"submitted_by":"/users/6800d05f-7213-48b1-9ad8-254c73c5b83f/"}
```
danxmoran: About `cross_references`:

It looks like this is actually following the same structure we use in ClinVar (assuming `GEO` is like a DB and `SAMN...` is like an ID). Would there be any value in aligning the struct formats?

danxmoran: About `age`: Looks like it should be parse-able as a number after all!

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -0,0 +1,59 @@
...
+      timeCreated = donorInput.read[OffsetDateTime]("date_created"),
+      age = donorInput.tryRead[String]("age"),
+      ethnicities = Array() ++ donorInput.tryRead[String]("ethnicity"),
+      organism = donorInput.read[String]("organism"),
+      sex = donorInput.read[String]("sex"),
+      award = donorInput.read[String]("award"),
+      lab = donorInput.read[String]("lab"),
+      lifeStage = donorInput.tryRead[String]("life_stage"),
+      parentIds = donorInput.read[Array[String]]("parents"),
+      status = donorInput.read[String]("status"),
```
danxmoran: Didn't we decide to drop `status` across the board?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  def transformDonor(donorInput: Msg): HumanDonor = {
+    HumanDonor(
+      id = donorInput.read[String]("accession"),
+      crossReferences = donorInput.read[Array[String]]("dbxrefs"),
+      timeCreated = donorInput.read[OffsetDateTime]("date_created"),
+      age = donorInput.tryRead[String]("age"),
+      ethnicities = Array() ++ donorInput.tryRead[String]("ethnicity"),
+      organism = donorInput.read[String]("organism"),
```
danxmoran: Do we need `organism` when the table includes human in the name? (Its presence or absence won't directly affect cross-dataset search)

transformation/src/test/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+      actual should contain theSameElementsAs expected
+    }
+  }
+
+  private val outputDirs = Set(
+    "human_donor"
+  )
+
+  outputDirs.foreach {
+    it should behave like compareTruthAndCompSets(_)
```
quazi-broad: Yay testing! So these two records in /test/test-files/inputs/ are for the unit test, right?
I think I need to incorporate the comparison method you've implemented to the encode-extract integration test, too. Good work!

transformation/src/jade-schema/tables/experiment_activity.table.json
```diff
@@ -89,11 +89,6 @@
       "datatype": "string",
       "type": "required"
     },
-    {
-      "name": "status",
-      "datatype": "string",
-      "type": "required"
-    },
```
snwessel: Dan made a good point that the status field should have been removed from all of the tables last week (I had only removed it from the files). This change is a bit tangential to the rest of this PR, but I am including it here so that it gets merged before other Encode transformation tickets are started.

transformation/src/jade-schema/tables/human_donor.table.json
snwessel: Good catch! We do definitely need to store age unit information separately. I added the `age_unit` column and made `age` an integer.

transformation/src/test/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineSpec.scala
snwessel: Yes, they are! I was mostly able to base this off of the existing ClinVar unit tests, but it's definitely a helpful tool and was really easy to use.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
snwessel: Yes, Encode is storing ethnicity as a single string but I believe other datasets store it as a list of strings. Since this falls into the realm of DSPCore, I think the idea is to choose a datatype that would work for multiple datasets.  @kreinold would you be able to chime in here and add anything I left out?

I believe the same would be said for the twin_ids field below.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
snwessel: Same as above.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
snwessel: I talked to Kathy about this and she said that she feels strongly that this will be important for cross dataset search. Maybe a thing to bring up with her if you disagree?

transformation/src/test/test-files/outputs/human_donor/part1.json
snwessel: @raaidbroad or @quazi-broad do either of you have thoughts or opinions about how we store cross references? Would it make more sense to store them in a struct like we do for clinvar? In Encode, the schema is currently just storing them as strings.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
snwessel: Yes, good catch. It seems like I had only removed them from Files before. I just removed them from the rest of the tables.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
snwessel: Age unit is actually stored as a separate field in Encode so it's easy to parse here 😁

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: This is similar to storing "status". Any hard-coded constants can be injected into the search documents at indexing time, saving us $$$ in storage. I'll talk with Kathy.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: Agreed that it is good to align column names and data types (and even data values when possible!) of the DSP Core fields, but I don't think it's worth bending the data to align the BigQuery column modes. Arrays have extra storage overhead, and they're more difficult to work with in SQL. When we index for search we'll be targeting Elasticsearch where everything is an array.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
snwessel: Just changed this to an optional string instead of an array.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
snwessel: Oh, good to know! I hadn't realized that Elasticsearch stored things differently. I just updated it to store as an optional string instead of an array.

transformation/src/test/test-files/outputs/human_donor/part1.json
danxmoran: We can leave it as-is for now; the datatype of the column is a string either way, so if we decide to switch later we can do so without a schema migration.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: Talked with Kathy. She thinks there's a high probability that we'll eventually be asked to ingest donor types other than "human". The least invasive way we can future-proof for that is:
1. Leave this column in here, with the hard-coded value
2. Rename the `human_donor` table to just `donor`

snwessel: Sounds good. I just pushed the changes to rename the table.

PR #20, opened 2020-03-16
Sw dspdc 893 use updated jade dataset plugin

snwessel:
[Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-893)

Bumping the version of the MonsterJadeDatasetPlugin, which required splitting the schema into its own sbt project.


project/plugins.sbt
```diff
@@ -11,7 +11,7 @@ resolvers += Resolver.url(
   new URL("https://broadinstitute.jfrog.io/broadinstitute/libs-release/")
 )(publishPatterns)
 
-val sbtPluginsVersion = "0.9.0"
+val sbtPluginsVersion = "0.12.0"
```
quazi-broad: I noticed this value in the clinvar project is "0.12.1"
https://github.com/DataBiosphere/clinvar-ingest/blob/master/project/plugins.sbt#L1
Should the version numbers be the same?

snwessel: Oh good catch! It probably should. @danxmoran do you agree?

danxmoran: Yes! Great catch Quazi, 0.12.0 is broken

PR #21, opened 2020-03-18
Fix warnings I noticed in ENCODE.

danxmoran:


extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -218,10 +218,12 @@ class ExtractionPipelineBuilder(getClient: () => EncodeClient)
   ): SCollection[String] => SCollection[Msg] = { idStream =>
     val paramsBatchStream =
       idStream
-        .map(KV.of("key", _))
+        .withName("Apply fake keys")
+        .map(KV.of("", _))
         .setCoder(KvCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of()))
         .applyKvTransform(GroupIntoBatches.ofSize(batchSize))
-        .map(_.getValue)
+        .withName("Strip away fake keys")
```
danxmoran: The problem here was that multiple `map` calls all nest within the same scope, resulting in non-unique generated names. Fixed by giving names to 2 of the 3 (which double as comments for why we're doing what we're doing)

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -12,6 +12,12 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
   /** (De)serializer for the upack messages we read from storage. */
   implicit val msgCoder: Coder[Msg] = Coder.beam(new UpackMsgCoder)
 
+  /** (De)serializer for the ODTs we extract from raw data. */
+  implicit val odtCoder: Coder[OffsetDateTime] = Coder.xmap(Coder.stringCoder)(
+    OffsetDateTime.parse(_),
+    _.toString
+  )
```
danxmoran: Added an explicit coder to prevent falling back to Kryo, which results in an warning.

transformation/src/test/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineSpec.scala
```diff
@@ -10,11 +10,16 @@ class TransformationPipelineSpec extends PipelineBuilderSpec[Args] {
...
   private val compareDirString = compareDir.pathAsString
-  private val inputDirString = s"${testFileLocation}/inputs"
+  private val inputDirString = s"$testFileLocation/inputs"
 
   override val testArgs =
     Args(inputPrefix = inputDirString, outputPrefix = compareDirString)
 
+  override def afterAll(): Unit = {
+    compareDir.delete()
+    ()
```
danxmoran: Make sure test outputs get cleaned up.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
quazi-broad: I was trying to read more about the withName function and found this definition:
Return a Value from this Enumeration whose name matches the argument s.
Is that how it's being used here? Do we apply those strings "Apply fake keys" and "Strip away fake keys" elsewhere so that they return when these calls are made?

danxmoran: That's a different withName. This one sets the name of the next transformation step in the pipeline (vs. using an auto-generated name).

PR #23, opened 2020-03-19
Sw dspdc 839 transform antibodies

snwessel:
Extending the Encode transformation to map Antibody objects ([Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-839)). A few notes below:

The `target` field has a slightly more complicated mapping than other fields. For example:
 - `["/targets/NIPBL-human/"]` should become `"NIPBL"`
 - `["/targets/H3K4me3-human/", "/targets/H3K4me3-mouse/"]` should become `"H3K4me3"`
 - `["/targets/FLAG-synthetic_tag/"]` should become `None` (synthetic flag type targets should be removed)
- I also included more test examples than usual to make sure that this mapping happens correctly.

These changes also included making some small tweaks to the Antibody table in the schema:
- `isotype` is no longer a required field
- `purification` is now stored as a list to match the format of the incoming encode data
- `target` is no longer a required field


transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -33,11 +33,9 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
     // read in extracted info
     val donorInputs = StorageIO
-      .readJsonLists(
-        ctx,
-        "Donors",
-        s"${args.inputPrefix}/Donor/*.json"
-      )
+      .readJsonLists(ctx, "Donors", s"${args.inputPrefix}/Donor/*.json")
+    val antibodyInputs = StorageIO
+      .readJsonLists(ctx, "AntibodyLot", s"${args.inputPrefix}/AntibodyLot/*.json")
```
danxmoran:
```suggestion
      .readJsonLists(ctx, "AntibodyLots", s"${args.inputPrefix}/AntibodyLot/*.json")
```
For consistency.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -103,4 +103,30 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
   }
+
+  def transformAntibody(antibodyInput: Msg): Antibody = {
+    // Use regular expressions to remove everything but the actual target name, which should be the same across
+    // all targets in the list. Remove "synthetic_tag" type targets.
+    val mappedTarget = antibodyInput
+      .read[Array[String]]("targets")
+      .filterNot(target => target.matches(".*synthetic_tag/"))
+      .headOption
+      .map(target => target.replaceAll(raw"\/targets\/|-[a-zA-Z]*\/", ""))
```
danxmoran: There's a slightly cleaner / more efficient way to do this.

First, you'll want to define a `val` outside of the method:
```scala
val EncodeIdPattern = "/[^/]+/([^/]+)/".r
```
Then this `map` can instead be a `collect`:
```suggestion
      .collect { case EncodeIdPattern(accession) => accession }
```

danxmoran: I would also flip the `collect` and the `filter`, so the filter can then be a little simpler:
```scala
.filterNot(_.endsWith("synthetic_tag"))
```

raaidbroad: whoa fancy

danxmoran: Oh I missed that this also needs to strip a suffix from the accession, hrm... I still like the idea of using the built-in regex matching capabilities, you'll just need to adjust the post-match logic. It can also double as the filter. Something like:
```
stuff
  .collect { moreStuff }
  .flatMap { accession =>
    val (front, back) = accession.split('-')
    if (back.endsWith(thing)) None
    else Some(front)
  }
```

transformation/src/test/test-files/inputs/AntibodyLot/part1.json
```diff
@@ -0,0 +1,4 @@
...140BWE/} does not have AR dbxrefs.","category":"missing antibody registry reference","level_name":"INTERNAL_ACTION","level":30,"path":"/antibodies/ENCAB140BWE/","name":"audit_antibody_dbxrefs_ar"}]},"award":"/awards/UM1HG009390/","characterizations":["/antibody-characterizations/7c13630f-d583-4979-b604-33f401842949/","/antibody-characterizations/5ede8c39-68d4-4d7d-b121-d94e98bb9a28/"],"clonality":"monoclonal","date_created":"2017-12-06T21:08:41.675624+00:00","dbxrefs":[],"host_organism":"/organisms/rabbit/","lab":"/labs/bradley-bernstein/","lot_id":"9","lot_id_alias":[],"lot_reviews":[{"detail":"Fully characterized.","status":"characterized to standards","biosample_term_name":"any cell type or tissue","targets":["/targets/H3K4me3-human/"],"organisms":["/organisms/human/"],"biosample_term_id":"NTR:99999999"}],"product_id":"9751BC","purifications":[],"schema_version":"9","source":"/sources/cell-signaling/","status":"released","submitted_by":"/users/f1843c60-e027-4b18-8582-64d3f3eae45b/","targets":["/targets/H3K4me3-human/"],"title":"ENCAB140BWE","url":"http://www.cellsignal.com/products/9751.html","used_by_biosample_characterizations":[],"uuid":"46aeb66a-d983-4b6f-ae1c-6aaa86960eff"}
```
danxmoran: This line has an audit value, should we add columns to this table to capture it? Ok to leave for a separate ticket (I've got a generic method for summarizing audits on my file branch).

transformation/src/test/test-files/inputs/AntibodyLot/part1.json
```diff
@@ -0,0 +1,4 @@
...
...nisms/mouse/","isotype":"IgG1","lab":"/labs/richard-myers/","lot_id":"SLBK1346V","lot_id_alias":[],"lot_reviews":[{"organisms":["/organisms/human/"],"detail":"Fully characterized.","biosample_term_id":"EFO:0001187","biosample_term_name":"HepG2","status":"characterized to standards","targets":["/targets/FLAG-synthetic_tag/"]},{"organisms":["/organisms/human/"],"detail":"Fully characterized.","biosample_term_id":"EFO:0001203","biosample_term_name":"MCF-7","status":"characterized to standards","targets":["/targets/FLAG-synthetic_tag/"]},{"organisms":["/organisms/human/"],"detail":"Fully characterized.","biosample_term_id":"EFO:0002067","biosample_term_name":"K562","status":"characterized to standards","targets":["/targets/FLAG-synthetic_tag/"]}],"product_id":"F1804","purifications":["affinity"],"schema_version":"9","source":"/sources/sigma/","status":"released","submitted_by":"/users/28bb3427-86e6-480a-be97-922c8b489f6a/","targets":["/targets/FLAG-synthetic_tag/","/targets/3xFLAG-synthetic_tag/"],"title":"ENCAB697XQW","url":"http://www.sigmaaldrich.com/catalog/product/sigma/f1804?lang=en&region=US","used_by_biosample_characterizations":[],"uuid":"534171de-26f6-4363-8a1e-a353357daea9"}
```
danxmoran: (Commenting on random line) Can we add a test case w/ multiple values in the targets array?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -103,4 +105,33 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
       submittedBy = donorInput.read[String]("submitted_by")
     )
   }
+
+  def transformAntibody(antibodyInput: Msg): Antibody = {
+    // Use regular expressions to remove everything but the actual target name, which should be the same across
+    // all targets in the list. Remove "synthetic_tag" type targets.
+    val mappedTarget = antibodyInput
+      .read[Array[String]]("targets")
+      .collectFirst { case EncodeIdPattern(accession) => accession } // filter out non-matches
```
danxmoran: Instead of `collectFirst` here, should we collect everything and make sure they all agree? A mismatch would be a case where logging a warning is appropriate.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -33,11 +35,9 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
     // read in extracted info
     val donorInputs = StorageIO
-      .readJsonLists(
-        ctx,
-        "Donors",
-        s"${args.inputPrefix}/Donor/*.json"
-      )
+      .readJsonLists(ctx, "Donors", s"${args.inputPrefix}/Donor/*.json")
+    val antibodyInputs = StorageIO
+      .readJsonLists(ctx, "AntibodyLots", s"${args.inputPrefix}/AntibodyLot/*.json")
```
danxmoran: Sorry to sneak in merge conflicts 🙁 there should be a new helper method that you can use to read these while stripping out unknown values.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -103,4 +105,33 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
       submittedBy = donorInput.read[String]("submitted_by")
     )
   }
+
+  def transformAntibody(antibodyInput: Msg): Antibody = {
```
danxmoran: I've started moving these methods into their own objects, could you do the same?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: There is now a `CommonTransformations.readId` method which might be useful here

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -103,4 +105,33 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+      .read[Array[String]]("targets")
+      .collectFirst { case EncodeIdPattern(accession) => accession } // filter out non-matches
+      .flatMap { accession =>
+        val (front, back) = accession.splitAt(accession.lastIndexOf('-'))
+        if (back == "-synthetic_tag") None
+        else Some(front)
+      }
+
+    Antibody(
+      id = antibodyInput.read[String]("accession"),
```
danxmoran: I would use `CommonTransformations.readId` here. There were a couple of files with no accession, and I'm wary that it might not be restricted to that object type.

transformation/src/test/test-files/inputs/AntibodyLot/part1.json
snwessel: Note: We talked about this over slack and decided not to store the audit values for antibodies.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/AntibodyTransformations.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import java.time.OffsetDateTime
+import org.broadinstitute.monster.encode.jadeschema.table.Antibody
+import org.slf4j.LoggerFactory
+import upack.Msg
+
+object AntibodyTransformations {
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  val EncodeIdPattern = "/[^/]+/([^/]+)/".r
```
danxmoran: This isn't needed here any more, it's included in `CommonTransformations`

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/AntibodyTransformations.scala
```diff
@@ -0,0 +1,55 @@
...
+  def transformAntibody(antibodyInput: Msg): Antibody = {
+    import org.broadinstitute.monster.common.msg.MsgOps
+
+    val id = antibodyInput.read[String]("accession")
+    val targets = antibodyInput
+      .read[Array[String]]("targets")
+
+    // Use regular expressions to remove everything but the actual target name, which should be the same across
+    // all targets in the list. Remove "synthetic_tag" type targets.
+    val mappedTargets = targets.collect {
```
danxmoran: This `collect` can become a `map`:
```
targets.map(CommonTransformations.readId)
```
We lose the filter-if-it-doesn't-match behavior, but that's probably a good thing; if we get malformed target strings, we probably want to go boom.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/AntibodyTransformations.scala
```diff
@@ -0,0 +1,55 @@
...
+    } // filter out non-matches
+    .flatMap { target =>
+      val (front, back) = target.splitAt(target.lastIndexOf('-'))
+      if (back == "-synthetic_tag") None
+      else Some(front)
+    }
+    val firstMappedTarget = mappedTargets.headOption
+
+    // check that all other target values match the first target value
+    if (!mappedTargets.forall(target => target == firstMappedTarget)) {
```
danxmoran: I don't think this equality check will work as-is. `target` has type `String`, but `firstMappedTarget` has type `Option[String]`.

You can simplify the whole thing by rearranging slightly. Add a `.toSet[String]` after the `flatMap`, at the end of the `mappedTargets` definition. The Set builder will de-duplicate values. Then you can:
1. Check if size > 1, and log the warning if so.
2. Take the `headOption` of the set for use in the output model

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/AntibodyTransformations.scala
snwessel: Would the `readId` method specifically be applicable here? It looks like it takes in a message and searches for the `@id` field, whereas we want to just read in a string (like `"/targets/.../"`).

I think the `transformId` method would be more fitting, right?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -37,17 +37,20 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
 
     // write back to storage
+    StorageIO.writeJsonLists(donorOutput, "Donors", s"${args.outputPrefix}/donor")
     StorageIO.writeJsonLists(
-      donorOutput,
-      "Donors",
-      s"${args.outputPrefix}/donor"
+      antibodyOutput,
+      "Antibodies",
+      s"${args.outputPrefix}/antibody"
```
quazi-broad: Smallest nit/question about style - is it preferable to have the params for a function all on one line vs on separate new lines? I noticed my IntelliJ likes to spread things out but it doesn't do it all the time so some similar calls in my code are formatted differently just like this.

snwessel: Ope, sorry. I merged before I saw your comment. I think the compiler reformats things when it runs, but I don't know all of the rules. It seems like this antibody output command exceeds the line width limit so it reformats it, but the donor command is slightly shorter.

snwessel: It would probably be worth making them all use the same format though, I agree.

PR #24, opened 2020-03-20
DSPDC-920 Refactor extraction to avoid large responses

danxmoran:
With a bunch of general cleanups thrown in. I'll comment inline to explain.

### DIFF
--- a/extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilderIntegrationSpec.scala
+++ b/extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilderIntegrationSpec.scala
@@ -7,8 +7,6 @@ class ExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args]
   val outputDir = File.newTemporaryDirectory()
   override def afterAll(): Unit = outputDir.delete()
 
-  val testBatchSize: Long = 100
-
   val allParams: List[(String, String)] = {
     List(
       // Default Human Params
@@ -20,7 +18,7 @@ class ExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args]
 
   /// KICK OFF EXTRACTION
   override val testArgs =
-    Args(outputDir.pathAsString, testBatchSize, allParams)
+    Args(outputDir.pathAsString, allParams)
   override val builder = ExtractionPipeline.pipelineBuilder
 
   // Paths for entity files downloaded
--- a/extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilderIntegrationSpec.scala
+++ b/extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilderIntegrationSpec.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.monster.encode.extraction
 import better.files.File
 import org.broadinstitute.monster.common.PipelineBuilderSpec
 
-class ExtractionPipelineIntegrationSpec extends PipelineBuilderSpec[Args] {
+class ExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args] {
   val outputDir = File.newTemporaryDirectory()
   override def afterAll(): Unit = outputDir.delete()
 
@@ -38,7 +38,7 @@ class ExtractionPipelineIntegrationSpec extends PipelineBuilderSpec[Args] {
   val targetPath = (outputDir.pathAsString + "/Target")
 
   /// VERIFY DOWNLOADS
-  behavior of "ExtractionPipeline"
+  behavior of "ExtractionPipelineBuilder"
 
   it should "successfully download AnalysisStep files" in {
     analysisStepPath.nonEmpty should be(true)
--- a/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
+++ b/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
@@ -20,10 +20,6 @@ case class Args(
   )
   outputDir: String,
   @HelpMessage(
-    "Batch size that defines the number of elements in a batch when making certain API calls"
-  )
-  batchSize: Long = 64L,
-  @HelpMessage(
     "Initial query to target a specific entry-point to the pipeline."
   )
   initialQuery: List[(String, String)] = List("organism.name" -> "human")
--- a/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
+++ b/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
@@ -22,7 +22,7 @@ case class Args(
   @HelpMessage(
     "Batch size that defines the number of elements in a batch when making certain API calls"
   )
-  batchSize: Long,
+  batchSize: Long = 64L,
   @HelpMessage(
     "Initial query to target a specific entry-point to the pipeline."
   )
--- a/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
+++ b/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
@@ -19,7 +19,7 @@ import scala.concurrent.Future
   *
   * @param getClient function that will produce a client which can interact with the ENCODE API
   */
-class ExtractionPipelineBuilder(getClient: () => EncodeClient)
+class ExtractionPipelineBuilder(batchSize: Long, getClient: () => EncodeClient)
     extends PipelineBuilder[Args]
     with Serializable {
   import ExtractionPipelineBuilder._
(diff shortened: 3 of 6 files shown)


extraction/src/it/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -3,7 +3,7 @@ package org.broadinstitute.monster.encode.extraction
 import better.files.File
 import org.broadinstitute.monster.common.PipelineBuilderSpec
 
-class ExtractionPipelineIntegrationSpec extends PipelineBuilderSpec[Args] {
+class ExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args] {
```
danxmoran: Renamed to mirror the main class, to match convention.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
```diff
@@ -22,7 +22,7 @@ case class Args(
   @HelpMessage(
     "Batch size that defines the number of elements in a batch when making certain API calls"
   )
-  batchSize: Long,
+  batchSize: Long = 64L,
```
danxmoran: Things seem to chug along smoothly with this.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -24,52 +21,86 @@ import scala.concurrent.Future
...
     extends PipelineBuilder[Args]
     with Serializable {
+  import ExtractionPipelineBuilder._
+
   implicit val coder: Coder[Msg] = Coder.beam(new UpackMsgCoder)
 
+  /** Convenience alias to make signatures a little less nasty. */
+  private type LookupParams = (EncodeEntity, List[(String, String)])
+
+  private val lookupFn =
```
danxmoran: This used to be a `def` that would produce one lookup function per encode entity. Refactored to push the entity types into the `asyncLookup` args so we could enforce a global concurrency limit.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -24,52 +21,86 @@ import scala.concurrent.Future
...
+    encodeEntity: EncodeEntity,
+    filterBatches: SCollection[List[(String, String)]]
+  ): SCollection[Msg] =
+    filterBatches
+      .withName("Construct full query parameters")
+      .map(encodeEntity -> _)
+      .withName("Query ENCODE API")
+      .applyKvTransform(ParDo.of(lookupFn))
+      .withName("Extract result graph")
+      .flatMap(_.getValue.fold(throw _, _.read[Array[Msg]]("@graph")))
```
danxmoran: Moved this up to be closer to the `lookupFn` definition, and added `withName`s so each step is clear in the UI.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -24,52 +21,86 @@ import scala.concurrent.Future
...
-      linkingEntities: SCollection[Msg],
-      linkedField: String = "@id"
+      sourceEntityType: EncodeEntity,
+      sourceField: String,
+      sourceEntities: SCollection[Msg],
+      targetEntityType: EncodeEntity,
+      targetField: String
     ): SCollection[Msg] = {
-      linkingEntities.transform(
-        s"Extract ${matchingField} from ${entityToExtract.entryName} data"
```
danxmoran: This description was very misleading: it made it sound like the thing we were extracting was already present, and we were pulling some field out of it. Reworded below,

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -24,52 +21,86 @@ import scala.concurrent.Future
...
-        out
+      val description =
+        s"Query ${targetEntityType.entryName} data using ${sourceEntityType.entryName} objects"
+
+      val out = sourceEntities.transform(description) { entities =>
+        val joinValues = getDistinctValues(sourceField, entities)
+        val queryBatches = groupValues(args.batchSize, joinValues.map(targetField -> _))
+        getEntities(targetEntityType, queryBatches)
       }
+      writeJsonLists(
```
danxmoran: I moved this out of the `transform` above so the top-level graph shows it.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -79,166 +110,152 @@ class ExtractionPipelineBuilder(getClient: () => EncodeClient)
...
-      entityToExtract = EncodeEntity.File,
-      matchingField = "@id",
-      linkingEntities = SCollection.unionAll(List(experiments, fcExperiments)),
-      linkedField = "dataset"
+      // Fudging a little, but it's only used for logging so it's ok.
+      sourceEntityType = EncodeEntity.Experiment,
+      sourceField = "files",
+      sourceEntities = SCollection.unionAll(List(experiments, fcExperiments)),
+      targetEntityType = EncodeEntity.File,
+      targetField = "@id"
```
danxmoran: This was the original motivation: instead of pulling out the `@id`s of experiments and joining against `dataset` (returning around an order of magnitude more things than `batchSize` asked for), we pull out `files` and join against `@id`.

extraction/src/test/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilderSpec.scala
```diff
@@ -2,198 +2,115 @@ package org.broadinstitute.monster.encode.extraction
...
 
-import scala.collection.mutable
-
-object EncodeExtractionPipelineBuilderSpec {
+object ExtractionPipelineBuilderSpec {
   // biosamples
   val biosampleParams = ("organism.name", "human")
 
   val biosampleOut = Obj(
-    Str("@graph") -> new Arr(
```
danxmoran: Pushed this common boilerplate into the mock client.

extraction/src/test/scala/org/broadinstitute/monster/encode/extraction/MockEncodeClient.scala
```diff
@@ -1,25 +1,20 @@
...
 
 /** Mock interface for ENCODE API. */
 class MockEncodeClient(
-  responseMap: Map[(EncodeEntity, (String, String)), Msg]
+  responseMap: Map[(EncodeEntity, (String, String)), List[Msg]]
 ) extends EncodeClient {
   val recordedRequests = mutable.Set[(EncodeEntity, (String, String))]()
 
   override def get(entity: EncodeEntity, params: List[(String, String)]): Future[Msg] = {
-    val paramsToUse = params.head
```
danxmoran: This `.head` was only safe when we were using batch size = 1. I see we bumped to a size of 2 at some point, which is probably good to test anyways, so I reworked this to deal with any number of input params.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
```diff
@@ -19,10 +19,6 @@ case class Args(
     "Path to directory where the extracted raw ENCODE metadata should be written"
   )
   outputDir: String,
-  @HelpMessage(
-    "Batch size that defines the number of elements in a batch when making certain API calls"
-  )
-  batchSize: Long,
```
danxmoran: It doesn't really make sense to parameterize this here; there's a size that works well in prod, so it's better to hard-code it and avoid confusion.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -21,224 +18,232 @@ import scala.concurrent.Future
...
-      matchingField = "accession",
-      linkingEntities = libraries,
-      linkedField = "library.accession"
+    val files = extractLinkedEntities(
+      sourceEntityType = EncodeEntity.Library,
+      sourceField = "@id",
+      sourceEntities = libraries,
+      targetEntityType = EncodeEntity.File,
+      targetField = "replicate_libraries"
     )
```
danxmoran: Querying files using experiments was bad either way we tried it.
* Querying on `file.dataset = experiment.id` resulted in huge response payloads
* Querying on `file.id = experiment.files` resulted in too many individual HTTP requests

I found this `replicate_libraries` field on the file objects. It's a roll-up of the file's direct libraries (for sequence files) and the libraries of all `derived_from` files. Running this code in Dataflow finished in a few minutes vs. 30+ minutes in the previous implementation.

PR #25, opened 2020-03-23
DSPDC-922 Strip "unknown" values as an initial pre-processing step

danxmoran:
ENCODE does the most annoying thing when they don't have info for a required field. Instead of leaving it blank and admitting it's not really required, they stick `"unknown"` in as the value. They do this even if the normal type of the field is a number.

It's much better for BQ to let `NULL` be the representation of no data, so we strip unknowns from every input stream before doing any other transformations.


transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -85,6 +88,22 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
   }
 
+  /** Message value used by ENCODE in place of null. */
+  val UnknownValue: Msg = Str("unknown")
+
+  /**
+    * Process an arbitrary ENCODE object to strip out all
+    * values representing null / absence of data.
+    */
+  def removeUnknowns(rawObject: Msg): Msg = {
```
raaidbroad: this looks good to me; do you think it'd be good practice to make this extensible to add other things to strip out in the future (or pre-process in other ways here), or would that be overkill for now? Since this looks like it is set up to have just the one `UnknownValue` as opposed to a set of possible strings or something.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -85,6 +88,22 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
     ()
   }
 
+  /** Message value used by ENCODE in place of null. */
+  val UnknownValue: Msg = Str("unknown")
```
raaidbroad: Should the "U" at the start of `UnknownValue` be lower-case for variable naming convention?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: I think it'd be overkill for now, but if/when we hit the next use case we should parameterize this one instead of making another hard-coded variant.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
danxmoran: This aligns with the convention of using UpperCamelCase for constants. It's also required to be this for practical reasons, because if it's lower-cased then you can't match on it as an equality check on the LHS of a `match` statement (if it's lower-cased, the parser views it as a new name generated by the pattern-match, not a reference to a constant).

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/EncodeTransformationPipelineBuilder.scala
```diff
@@ -31,20 +36,18 @@ object EncodeTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+        .map(removeUnknowns)
 
-    val fileInputs = StorageIO
-      .readJsonLists(
-        ctx,
-        "Files",
-        s"${args.inputPrefix}/File/*.json"
-      )
+    // read in extracted info
+    val donorInputs = readRawEntities(EncodeEntity.Donor)
```
quazi-broad: Nice, this looks a lot neater!

PR #27, opened 2020-03-24
Add file transformations

danxmoran:
Including:
* DSPDC-899
* DSPDC-841
* DSPDC-842
* DSPDC-843

I'll comment inline with explanations / big lingering questions. The biggest remaining TODO is implementing the data modality check.


schema/src/main/jade-structs/read_structure.struct.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "read_structure",
```
danxmoran: I found that read structure is a structured object in ENCODE's raw file schema. This is a mirror of that structure. It might be overkill to parse->serialize back to string; we could instead stringify the raw JSON objects and achieve the same results with less chance of things breaking. Thoughts?

schema/src/main/jade-tables/alignment_file.table.json
```diff
@@ -11,11 +11,6 @@
       "datatype": "string",
       "type": "repeated"
     },
-    {
-      "name": "byte_size",
```
danxmoran: Removed the size column across the board since it's tracked for free by GCS (and doesn't seem like a high-level search criteria, though I could be wrong)

schema/src/main/jade-tables/alignment_file.table.json
```diff
@@ -33,16 +28,15 @@
...
-      "type": "required"
+      "datatype": "string"
     },
     {
       "name": "award",
       "datatype": "string",
       "type": "required"
     },
     {
-      "name": "file_format_type",
```
danxmoran: file_format_type only has values for a subset of data types, all of which get categorized as "other"

schema/src/main/jade-tables/alignment_file.table.json
```diff
@@ -60,23 +54,13 @@
       "datatype": "string",
       "type": "repeated"
     },
-    {
-      "name": "restriction_enzymes",
```
danxmoran: Only defined for a specific data type, which gets categorized as "other"

schema/src/main/jade-tables/sequence_file.table.json
```diff
@@ -60,89 +58,61 @@
...
-      "datatype": "string",
-      "type": "repeated"
-    },
     {
       "name": "submitted_by",
       "datatype": "string",
       "type": "required"
     },
     {
-      "name": "genome_annotation",
```
danxmoran: Note to self: double-check genome annotation for all categories

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/FileTransformations.scala
```diff
@@ -0,0 +1,267 @@
...
+      .unzip3
+
+    FileBranches(seq.flatten, align.flatten, other.flatten)
+  }
+
+  /** Compute the data modality of a raw file. */
+  private def computeDataModality(rawFile: Msg): String = {
+    rawFile match {
+      // FIXME: Actually compute something here once we know the formula.
+      case _ => "I'm a modality!"
```
danxmoran: Note to self: Implement this, or make a ticket to do so.

transformation/src/test/test-files/inputs/File/part-1.json
```diff
@@ -1,4 +1,9 @@
...
...ataset":"/experiments/ENCSR467QPI/","date_created":"2016-10-11T21:05:58.194169+00:00","dbxrefs":["SRA:SRR020257"],"fastq_signature":["TEMP:3:1::@EVANS_3021:3:1:0:1061"],"file_format":"fastq","file_size":1545668172,"file_type":"fastq","flowcell_details":[],"href":"/files/ENCFF797REO/@@download/ENCFF797REO.fastq.gz","lab":"/labs/zhiping-weng/","library":"/libraries/ENCLB258GAJ/","md5sum":"345315de7073c03d09628e151a5e59bb","no_file_available":false,"output_category":"raw data","output_type":"reads","platform":"/platforms/OBI%3A0000703/","quality_metrics":[],"read_count":20481685,"read_length":87,"read_length_units":"nt","replicate":"/replicates/87dae9d3-f68e-4b26-a2b7-a3096f2e2c87/","replicate_libraries":["/libraries/ENCLB258GAJ/"],"run_type":"single-ended","s3_uri":"s3://encode-public/2016/10/11/87818cbf-d9b8-4afe-9ddf-14dde0cafa15/ENCFF797REO.fastq.gz","schema_version":"17","status":"released","submitted_by":"/users/5a1c0169-f9d0-4598-8f92-3eb84b122c1e/","submitted_file_name":"SRR020257.fastq.gz","superseded_by":[],"technical_replicates":["2_2"],"title":"ENCFF797REO","uuid":"87818cbf-d9b8-4afe-9ddf-14dde0cafa15","read_structure":[{"sequence_element":"fake","start":0,"end":1000000}]}
```
danxmoran: Note: I couldn't actually find any files with a defined `read_structure` in our extracted data, so I added a fake one based on ENCODE's published schema to this line.

schema/src/main/jade-structs/read_structure.struct.json
snwessel: It feels like the question is if Encode changes things, do we want it to throw an error during our transformation? Or do we want to just accept the new data format? If this is likely to be used in search, maybe we should opt for the consistent format.

Do you know if this something people would be likely to search by?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/FileTransformations.scala
danxmoran: https://broadinstitute.atlassian.net/browse/DSPDC-938

PR #28, opened 2020-03-25
DSPDC-845 Initialize experiment transformations.

danxmoran:
This is all we can do without having the biosample-replicate join working. Will make a separate ticket for the follow-up work.


schema/src/main/jade-tables/experiment_activity.table.json
```diff
@@ -17,14 +17,8 @@
       "type": "required"
     },
     {
-      "name": "time_started",
-      "datatype": "timestamp",
-      "type": "required"
-    },
-    {
-      "name": "time_ended",
```
danxmoran: As far as I could see in the data and the mapping sheet, ENCODE doesn't publish this information for experiments, so there's not much we can do.

PR #31, opened 2020-03-31
DSPDC-941 Bump dependencies.

danxmoran:
Most of the code changes are from the new Scalafmt settings bundled in the plugin upgrade.

### DIFF
--- a/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
+++ b/extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
@@ -5,15 +5,15 @@ import caseapp.core.Error.MalformedValue
 import caseapp.core.argparser.{ArgParser, SimpleArgParser}
 import org.broadinstitute.monster.EncodeExtractionBuildInfo
 
-@AppName("ENCODE Extraction Pipeline")
-@AppVersion(EncodeExtractionBuildInfo.version)
-@ProgName("org.broadinstitute.monster.etl.encode.ExtractionPipeline")
 /**
   * Command-line arguments for the ETL workflow.
   *
   * scio's `ContextAndArgs.typed` delegates down to `caseapp`, which will generate
   * parsers + help text for these args (as well as Beams' underlying options)
   */
+@AppName("ENCODE Extraction Pipeline")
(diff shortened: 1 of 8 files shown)


extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
```diff
@@ -5,15 +5,15 @@ import caseapp.core.Error.MalformedValue
...
-@ProgName("org.broadinstitute.monster.etl.encode.ExtractionPipeline")
 /**
   * Command-line arguments for the ETL workflow.
   *
   * scio's `ContextAndArgs.typed` delegates down to `caseapp`, which will generate
   * parsers + help text for these args (as well as Beams' underlying options)
   */
+@AppName("ENCODE Extraction Pipeline")
+@AppVersion(EncodeExtractionBuildInfo.version)
+@ProgName("org.broadinstitute.monster.etl.encode.ExtractionPipeline")
```
danxmoran: Not scalafmt: Putting the docstring under the annotations caused it to be marked as "unmoored", leading to a failure in packaging.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/ExtractionPipelineBuilder.scala
```diff
@@ -62,17 +62,17 @@ class ExtractionPipelineBuilder(batchSize: Long, getClient: () => EncodeClient)
       .flatMap(_.getValue.fold(throw _, _.read[Array[Msg]]("@graph")))
 
   override def buildPipeline(ctx: ScioContext, args: Args): Unit = {
-    /**
```
danxmoran: Not scalafmt: Scaladoc-style comments (double ** at the front) are only valid on top-level members in a class. If they're used on nested methods like this one, they're marked as "unmoored" and cause packaging to fail.

extraction/src/main/scala/org/broadinstitute/monster/encode/extraction/Args.scala
quazi-broad: good to know

PR #32, opened 2020-03-31
Sw dspdc 838 transform libraries

snwessel:
Adding the transformation for libraries.

Also removing a few fields from the library table:
- `source`, which is rarely used in encode 
- `product_id` and `lot_id` are both related to the purchase of biosamples, and would make more sense to store in the biosample table. I will create a follow-up ticket to add them there.


transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/LibraryTransformations.scala
```diff
@@ -0,0 +1,30 @@
...
+      id = CommonTransformations.readId(libraryInput),
+      crossReferences = libraryInput.read[Array[String]]("dbxrefs"),
+      timeCreated = libraryInput.read[OffsetDateTime]("date_created"),
+      award = libraryInput.read[String]("award"),
+      lab = libraryInput.read[String]("lab"),
+      queriedRnpSizeRange = libraryInput.tryRead[String]("queried_RNP_size_range"),
+      rnaIntegrityNumber = libraryInput.tryRead[Double]("rna_integrity_number"),
+      sizeRange = libraryInput.tryRead[String]("size_range"),
+      strandSpecificity = libraryInput.read[Boolean]("strand_specificity"),
+      treatments = libraryInput.read[Array[String]]("treatments"),
```
danxmoran: Do we trust that this array will always be populated in the source data?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/LibraryTransformations.scala
```diff
@@ -0,0 +1,30 @@
...
+      submittedBy = libraryInput.read[String]("submitted_by"),
+      spikeIns = libraryInput.read[Array[String]]("spikeins_used"),
+      biosampleId = libraryInput.read[String]("biosample")
```
danxmoran: This value needs transformed to strip the `/biosamples/` prefix for the join against the `biosample` table to work.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/LibraryTransformations.scala
snwessel: From what I've seen, the extracted json always has `"treatments": []` when there are no treatments. I haven't seen any cases where the field is not defined, if that's what you're asking. We could change this to a `tryRead`, but would need to update it elsewhere too (we are using `read` to extract arrays of strings in the other transformations).

danxmoran: I'm ok leaving it until we see a problem in practice

danxmoran:
Heads-up: merging master will pull in new scalafmt settings, you might hit a failure for malformatted code if you do it in the browser

PR #35, opened 2020-04-02
DSPDC-934 Update test inputs/outputs to be a more realistic subset

danxmoran:
I used the scripts in `hack/` to populate these new inputs / outputs. The inputs should be more realistically linked to one another now.

I had to make some tweaks to the library schema to deal with ENCODE changing a type out from under us 🤕


transformation/src/test/test-files/inputs/Target/part-1.json
```diff
@@ -0,0 +1,6 @@
...
+{"@id":"/targets/H3K4me3-human/","@type":["Target","Item"],"aliases":[],"dbxref":["GeneID:333932","GeneID:653604","GeneID:126961","UniProtKB:Q71DI3"],"genes":["/genes/333932/","/genes/653604/","/genes/126961/"],"investigated_as":["histone","narrow histone mark"],"label":"H3K4me3","modifications":[{"amino_acid_code":"K","modification":"Trimethylation","position":4}],"name":"H3K4me3-human","organism":"/organisms/human/","schema_version":"13","status":"released","title":"H3K4me3 (Homo sapiens)","uuid":"65b74663-aedd-46ae-b2c5-f3e37c834867"}
+{"@id":"/targets/H3K9me3-human/","@type":["Target","Item"],"aliases":[],"dbxref":["GeneID:333932","GeneID:653604","GeneID:126961","UniProtKB:Q71DI3"],"genes":["/genes/333932/","/genes/653604/","/genes/126961/"],"investigated_as":["histone","broad histone mark"],"label":"H3K9me3","modifications":[{"amino_acid_code":"K","modification":"Trimethylation","position":9}],"name":"H3K9me3-human","organism":"/organisms/human/","schema_version":"13","status":"released","title":"H3K9me3 (Homo sapiens)","uuid":"14173a5f-799e-4dae-b155-a142c5b6b566"}
```
snwessel, line 6: Given that we've just been getting target information from a field the Antibody (as opposed to joining with the target objects), do we still need to be extracting targets?

danxmoran: As implemented now I'd say no, but I think it'd be good for us to start using them. The ID-hacking we're doing in the antibodies "works", but it'd be a lot more reliable to join against the actual raw targets and filter / flatten based on their metadata. What do you think?

snwessel: Yeah, I agree. I would definitely feel better about doing a join than the regex thing we're doing now.

PR #37, opened 2020-04-02
DSPDC-939 Add joins needed to populate experiment biosamples.

danxmoran:
Surprisingly straightforward!

I sanity-checked the new outputs by making sure that all of the biosamples listed in our `hack` script end up associated with an experiment, and they do. I haven't done the deeper check of making sure the associations are correct (yet).


danxmoran:
Well, that's not reassuring. The test passes locally...

raaidbroad:
All the tests passed on my machine too

danxmoran:
Sleep tells me that it's probably the sort-order of the array

PR #39, opened 2020-04-09
Sw dspdc add product and lot ids to biosample

snwessel:
[Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-948)

Joining Libraries to Biosamples in order to populate the `product_id` and `lot_id` fields on the biosample table (both are found on the Library object in Encode).

We assume that every library associated with a biosample will have the same product_id and lot_id. I went through five or six examples manually, in addition to the test cases, to make sure that this actually holds true.


transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -3,15 +3,38 @@ package org.broadinstitute.monster.encode.transformation
...
+    if (product_ids.size > 1) {
+      logger.warn(
+        s"Biosample '$biosample_id' has more than one product id in: [${product_ids.mkString(",")}]."
+      )
+    }
+    if (product_ids.size > 1) {
+      logger.warn(
+        s"Biosample '$biosample_id' has more than one lot id in: [${lot_ids.mkString(",")}]."
+      )
+    }
```
danxmoran, line 36: Did you see this in any of our test inputs? We can run it on the full extraction if not, just to see if it ever happens

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -38,7 +61,9 @@ object BiosampleTransformations {
       replicationType = "repType",
       treatments = biosampleInput.read[Array[String]]("treatments"),
       wasPerturbed = biosampleInput.read[Boolean]("perturbed"),
-      submittedBy = biosampleInput.read[String]("submitted_by")
+      submittedBy = biosampleInput.read[String]("submitted_by"),
+      productId = product_ids.headOption,
+      lotId = lot_ids.headOption
```
danxmoran, line 66: Should we do `headOption` here (effectively choosing randomly) if there are > 1 IDs? Or should we leave it as `None` to mark an error state?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
snwessel: Yeah, it seems like every library associated with a given biosample has the same values for product and lot id. I checked that this holds true in the test cases and the other examples I looked through by hand, but I agree, it would be good to run it on the full extraction and see what happens.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
snwessel: I like that suggestion. I'll change it to store None when there's an error.

snwessel: I just made the change. Let me know if it's what you had in mind.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -3,15 +3,42 @@ package org.broadinstitute.monster.encode.transformation
...
+  def transformBiosample(biosampleInput: Msg, joinedLibraries: Iterable[Msg]): Biosample = {
+    val biosample_id = CommonTransformations.readId(biosampleInput)
     val (auditLevel, auditLabels) = CommonTransformations.summarizeAudits(biosampleInput)
+
+    var product_ids = joinedLibraries
+      .flatMap(library => library.tryRead[String]("product_id"))
+      .toSet[String]
+    var lot_ids = joinedLibraries
+      .flatMap(library => library.tryRead[String]("lot_id"))
+      .toSet[String]
```
danxmoran, line 25: These can both stay `val`s

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -3,15 +3,42 @@ package org.broadinstitute.monster.encode.transformation
...
+
+    // check that there are not more than one unique product_id or lot_id.
+    // If there are multiple, default to None.
+    if (product_ids.size > 1) {
```
danxmoran, line 29: You can push these `if`s down to the `Biosample.appy` call, and add `else` branches to make them expressions

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -38,7 +65,9 @@ object BiosampleTransformations {
       replicationType = "repType",
       treatments = biosampleInput.read[Array[String]]("treatments"),
       wasPerturbed = biosampleInput.read[Boolean]("perturbed"),
-      submittedBy = biosampleInput.read[String]("submitted_by")
+      submittedBy = biosampleInput.read[String]("submitted_by"),
+      productId = product_ids.headOption,
```
danxmoran, line 69: This would become:
```scala
productId = if (productIds.size > 1) {
  logger.warn("Stuff")
  None
} else {
  productIds.headOption
}
```

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
danxmoran: Also style: should use camelCase instead of snake_case for names

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -3,15 +3,42 @@ package org.broadinstitute.monster.encode.transformation
...
 /** Transformation logic for ENCODE biosample objects. */
 object BiosampleTransformations {
   import org.broadinstitute.monster.common.msg.MsgOps
 
+  private val logger = LoggerFactory.getLogger(getClass)
+
   /** Transform a raw ENCODE biosample into our preferred schema. */
-  def transformBiosample(biosampleInput: Msg): Biosample = {
+  def transformBiosample(biosampleInput: Msg, joinedLibraries: Iterable[Msg]): Biosample = {
+    val biosample_id = CommonTransformations.readId(biosampleInput)
```
danxmoran, line 17: CamelCase here too:
```suggestion
    val biosampleId = CommonTransformations.readId(biosampleInput)
```

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -38,7 +50,21 @@ object BiosampleTransformations {
...
+      submittedBy = biosampleInput.read[String]("submitted_by"),
+      productId = if (productIds.size > 1) {
+        logger.warn(
+          s"Biosample '$biosampleId' has multiple product ids: [${productIds.mkString(",")}]."
+        )
+        None
+      } else {
+        productIds.headOption
+      },
+      lotId = if (productIds.size > 1) {
```
danxmoran, line 62:
```suggestion
      lotId = if (lotIds.size > 1) {
```

snwessel: Ah, darn 🤦‍♀️, yes good catch. Thank you!

PR #40, opened 2020-04-09
DSPDC-955 Add BiosampleType

danxmoran:
This PR:
* Adds BiosampleType to our extracted entities
* Uses extracted BiosampleType objects to populate data in our transformed biosample rows
* Updates our test inputs and outputs


transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineBuilder.scala
```diff
@@ -148,23 +148,39 @@ object TransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+    val biosampleTypeInputs = readRawEntities(EncodeEntity.BiosampleType)
+
+    val typesById = biosampleTypeInputs
+      .withName("Key biosample types by ID")
+      .keyBy(_.read[String]("@id"))
+
+    val biosamplesWithTypes = biosampleInputs
+      .withName("Key biosamples by type")
+      .keyBy(_.read[String]("biosample_ontology"))
+      .join(typesById)
```
danxmoran, line 162: `join` is an inner join, so any biosamples w/o an ontology will be silently dropped. That _should_ be fine since we expect every sample to have a linked type, but I could see an argument for using an outer join here and having the transformation log if a `None` slips through

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
```diff
@@ -32,10 +36,8 @@ object BiosampleTransformations {
...
         biosampleInput.tryRead[String]("part_of").map(CommonTransformations.transformId),
-      // TODO needs biosampleType join
-      anatomicalSite = "site",
-      // TODO needs biosampleType join
-      biosampleType = "type",
+      anatomicalSite = joinedType.read[String]("term_id"),
+      biosampleType = joinedType.read[String]("classification"),
       samplePreservationState = biosampleInput.tryRead[String]("preservation_method"),
       seeAlso = biosampleInput.tryRead[String]("url"),
       donorId = CommonTransformations.transformId(biosampleInput.read[String]("donor")),
```
danxmoran, line 43: There is one more join lurking below (needs Experiment info). I'm going to knock it out while I'm in here.

danxmoran: Errr, actually not. It looks like another case where we could end up with many replication-related values for a single sample, will make a ticket for it.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineBuilder.scala
danxmoran: I have convinced myself that silently dropping data is bad, so this should be changed

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/BiosampleTransformations.scala
snwessel: Is this ready for review then?

danxmoran: Yup!

PR #42, opened 2020-04-13
Sw dspdc 959 use extracted targets

snwessel:
Joining using the list of target ids on the antibody object. Then using the target objects in the antibody transformation. 

Update: It works! 🎉


transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineBuilder.scala
```diff
@@ -36,11 +36,31 @@ object TransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+      .withName("Key targets by id")
+      .keyBy(_.read[String]("@id"))
+
+    // join antibodies (currently a messy process) TODO: clean up
+    val antibodyOutput = antibodyInputs.map { rawAntibody =>
+      rawAntibody
+        .tryRead[Array[String]]("targets")
+        .getOrElse(Array.empty)
+        .map((_, rawAntibody))
+    }.flatten
```
snwessel, line 52: The process of converting to an array of `(targetId, antibody)` tuples filters out any antibodies that don't have targets. I have been trying to come up with ways to get around this issue, but haven't had any luck so far:
- I tried making the target id an option in the tuple, but doing so causes an error in the join later on, which expects a string, not an option. 
- I thought about keeping track of antibodies that don't have targets and appending them back on at the end, but this is messy and doesn't work anyway since  scollections don't seem to support append/concat methods. 

Any suggestions are appreciated, even if that means rethinking the approach 🙂

snwessel: I feel less than satisfied with the way I'm doing this join right now, so any more general feedback is appreciated too.

danxmoran: It's a little more verbose, but I would separate out the processes of grouping & joining the targets. Start with the logic to get all the (antibody-id, target-id) pairs:
```scala
val antibodyTargetPairs = antibodyInputs.flatMap { rawAntibody =>
  rawAntibody.tryRead[Array[String]]("targets").getOrElse(Array.empty).map { targetId =>
    rawAntibody.read[String]("@id") -> targetId
  }
}
```
Then you can join in the raw targets, and re-group by antibodies:
```scala
val targetsByAntibody = antibodyTargetPairs
  .keyBy(_._2)
  .join(targetInputs)
  .values
  // I think the pseudo-code falls apart here, but the idea is to key by the antibody ID
  .keyBy(_._1)
  .groupByKey
```
Then you'll have an `SCollection[(String, Iterable[Msg])]`. You can `antibodyInputs.leftJoin` that collection, and you'll end up with `Option[Iterable[Msg]]` for the linked targets. `.toIterable.flatten` on that will strip away the `Option` wrapper

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/AntibodyTransformations.scala
```diff
@@ -8,19 +8,13 @@ import upack.Msg
...
-      .map(CommonTransformations.transformId)
-      .flatMap { target =>
-        val (front, back) = target.splitAt(target.lastIndexOf('-'))
-        if (back == "-human") Some(front)
-        else None
-      }
+    val targetNames = joinedTargets
+      .filter(_.read[String]("organism") == "/organisms/human/")
+      .map(_.read[String]("label"))
+      .toArray
```
danxmoran, line 17: This part LGTM!

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineBuilder.scala
snwessel: Thanks Dan! I agree that this approach ^ would work better.

snwessel: It works! In retrospect, I realize I had been making this a lot more complicated than it needed to be 😁 ... Let me know what you think of the changes.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/AntibodyTransformations.scala
danxmoran: Oh wait, sorry. This needs a `.sorted.distinct` to ensure reproducibility, since there's no guarantee on the order of `joinedTargets`

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/TransformationPipelineBuilder.scala
```diff
@@ -36,11 +36,47 @@ object TransformationPipelineBuilder extends PipelineBuilder[Args] {
...
-    // So can antibodies.
+    // The Antibody transformation needs information from the Target objects
     val antibodyInputs = readRawEntities(EncodeEntity.AntibodyLot)
+    val targetInputs = readRawEntities(EncodeEntity.Target)
+    val keyedTargets = targetInputs
+      .withName("Key targets by id")
+      .keyBy(_.read[String]("@id"))
+
+    // Get all of the (Antibody-id, Target-id) pairs
+    val antibodyTargetPairs = antibodyInputs.map { rawAntibody =>
```
danxmoran, line 47: Using a `flatMap` here would be more idiomatic than a `map` followed by `flatten`. It might be more efficient at runtime, too.

danxmoran: Can we also add a `withName` after `antibodyInputs` to describe the operation?

raaidbroad: +1 to using `flatMap`

snwessel: Just updated 👍

PR #43, opened 2020-04-13
QH - DSPDC-938 - Update file transformation to calculate real data modalities

quazi-broad:
Initial commit - Adding transformation logic for different values of "assay_term_name" which correspond to different data modalities as per the reference document. Will need to update the test file outputs to ensure that the unit tests complete successfully with the new extended outputs.

Notes:
I may be using the wrong field to figure out the data modality. The reference document links "assay_title" values to data modality. When I was searching the test input files I found a lot of assay_title values within another field (/File/assay_term_name) so I thought that was the same as "assay_title". Turns out there are a lot of values within "assay_term_name" that aren't linked in the reference document.

I found fields labeled "assay_title" within the /Experiment and /FunctionalCharacterizationExperiment input files so I will have to refactor the code to pull from there to computer the data modality.

If there are cases where there's an "assay_title" value that isn't mapped, should we use empty string as the default (case _ => "" )?


transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/FileTransformations.scala
```diff
@@ -98,9 +98,74 @@ object FileTransformations {
 
   /** Compute the data modality of a raw file. */
   private def computeDataModality(rawFile: Msg): String =
-    rawFile match {
-      // FIXME: Actually compute something here once we know the formula.
-      case _ => "I'm a modality!"
+    rawFile.tryRead[String]("assay_term_name") match {
+      case None                            => ""
```
danxmoran, line 102: If there's ever a case where we can't determine the correct modality, we should output `None` instead of a placeholder value. You'll need to update the schemas of the file tables to remove the `required` modifier on each column.

danxmoran: This can then become slightly less verbose:
```scala
rawFile.tryRead[String]("assay_term_name").collect {
  case "3' RACE" => "Transcriptomic"
  .. etc ...
}
```

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/FileTransformations.scala
```diff
@@ -98,9 +98,74 @@ object FileTransformations {
...
+      case Some("scRNA-seq")            => "Transcriptomic_Unbiased"
+      case Some("single-cell ATAC-seq") => "Epigenomic_DNAChromatinAccessibility"
+      case Some("snATAC-seq")           => "Epigenomic_DNAChromatinAccessibility"
+      case Some("siRNA RNA-seq")        => "Transcriptomic_Unbiased"
+      case Some("small RNA-seq")        => "Transcriptomic_Unbiased"
+      case Some("Switchgear")           => "Epigenomic_RNABinding"
+      case Some("TAB-seq")              => "Epigenomic_DNAMethylation"
+      case Some("TF ChIP-seq")          => "Epigenomic_DNABinding_TranscriptomeFactorLocation"
+      case Some("total RNA-seq")        => "Transcriptomic_Unbiased"
+      //case Some("RNA microarray")          => "Transcriptomic_Unbiased"
```
danxmoran, line 165: Have you talked with Kathy / Sarah about what to do for this duplicate case?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/FileTransformations.scala
```diff
@@ -31,16 +31,18 @@ object FileTransformations {
   private val logger = LoggerFactory.getLogger(getClass)
 
   /** Divide a stream of raw files into branches, with one branch per file category. */
-  def partitionRawFiles(rawStream: SCollection[Msg]): FileBranches[SCollection, Msg] = {
+  def partitionRawFiles(
+    rawStream: SCollection[(Msg, Option[Msg])]
```
snwessel, line 35: It feels a little weird to be passing the experiments into this method where they're not being used, but I see why you are doing it since joining the experiments later on would be messy.

danxmoran: If we add any more JOINs (i.e. for pipelines), I think the move will be to declare a new `case class` that wraps up the raw file and all of its linked metadata, and pass those around through these methods. For now this is OK.

PR #44, opened 2020-04-14
DSPDC-846 Add library_prep generation to transformation pipeline.

danxmoran:


schema/src/main/jade-tables/library_prep_activity.table.json
```diff
@@ -6,35 +6,13 @@
       "datatype": "string",
       "type": "primary_key"
     },
-    {
-      "name": "cross_references",
-      "datatype": "string",
-      "type": "repeated"
-    },
-    {
-      "name": "time_created",
```
danxmoran, line 15: `time_created` and `cross_references` could both be pulled from the raw library, but I don't see much of a point in duplicating the info across tables.

snwessel: That makes sense to me.

transformation/src/test/test-files/outputs/library_prep_activity/part-1.json
```diff
@@ -0,0 +1,134 @@
+{"biosample_id":"ENCBS051RNA","id":"ENCLB179ZZZ","library_id":"ENCLB179ZZZ","library_material":"SO:0000356","library_material_name":"RNA"}
```
danxmoran, line 1: Duplicating the library ID here is kinda weird / wasteful. We _could_ drop that column and just use the `library_id` as the PK.

I'm also unsure of the general usefulness of this table vs. just adding the two extra columns to the `library` table. I get that conceptually there are 2 different things being stored in 2 different tables, but it's not clear to me if ENCODE exposes enough info about the activity for it to be useful as a separate table.

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/LibraryTransformations.scala
```diff
@@ -36,4 +37,17 @@ object LibraryTransformations {
...
   }
+
+  /** Transform a raw ENCODE library into our preferred library-prep schema. */
+  def transformLibraryPrep(libraryInput: Msg): LibraryPrepActivity = {
+    val id = CommonTransformations.readId(libraryInput)
+    LibraryPrepActivity(
+      id = id,
+      libraryMaterial = libraryInput.tryRead[String]("nucleic_acid_term_id"),
+      libraryMaterialName = libraryInput.tryRead[String]("nucleic_acid_term_name"),
+      libraryId = id,
```
snwessel, line 48: Having two columns with the same value feels weird, but I don't know how we would combine them and/or remove one, without just combining the Library Prep table with the Library table. 

Are we keeping them separate because they will be updated in different ways? I hadn't realized that there was a 1-1 relationship between them, so I want to make sure I understand the reasoning behind separating them.

transformation/src/test/test-files/outputs/library_prep_activity/part-1.json
snwessel: Oops, I missed this comment when I made my comment below...

snwessel: I definitely agree that it might make sense to combine them... Maybe a good thing to get Kathy's thoughts on if we meet up to talk about Encode tomorrow morning?

transformation/src/main/scala/org/broadinstitute/monster/encode/transformation/LibraryTransformations.scala
danxmoran: I have similar confusion. I think that having separate tables is the most literal translation of the data model, but it might not make sense for ENCODE.

PR #47, opened 2020-04-17
Sw dspdc 900 create orchestration workflow

snwessel:
Setting up an orchestration workflow for Encode ([Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-900)).

* Related to this PR in monster-deploy: https://github.com/broadinstitute/monster-deploy/pull/60


orchestration/values.yaml
```diff
@@ -0,0 +1,17 @@
...
+  storageClass: standard
+  downloadSize: 1Gi
+  extractSize: 10Gi
+cron:
+  enable: true
+  # Every Sunday at 3:30am
+  schedule: '30 3 * * Sun'
+  timezone: 'America/New_York'
+dataflow:
+  region: us-west-1
```
snwessel, line 11: I updated the region, but the rest of these values were copied from https://github.com/DataBiosphere/clinvar-ingest/blob/master/orchestration/values.yaml

orchestration/values.schema.json
```diff
@@ -0,0 +1,70 @@
...
+        "useFlexRS": { "type": "boolean" },
+        "pipelineVersion": { "type": "string" }
+      },
+      "required": ["project", "region", "tmpBucketName",
+        "subnetName", "workerAccount", "workerMachineType",
+        "autoscaling", "useFlexRS", "pipelineVersion"]
+      }
+    },
+  "required": ["gcs", "serviceAccount", "volumes", "cron", "dataflow"]
+}
```
snwessel, line 70: Copied from https://github.com/DataBiosphere/clinvar-ingest/blob/master/orchestration/values.schema.json

orchestration/templates/run-transformation-pipeline.yaml
```diff
@@ -0,0 +1,39 @@
...
+          {{- with .autoscaling }}
+          - --autoscalingAlgorithm=THROUGHPUT_BASED
+          - --numWorkers={{ .minWorkers }}
+          - --maxNumWorkers={{ .maxWorkers }}
+          {{- end }}
+          {{- if .useFlexRS }}
+          - --flexRSGoal=COST_OPTIMIZED
+          {{- else }}
+          - --experiments=shuffle_mode=service
+          {{- end }}
```
snwessel, line 38: The args here are copied from the clinvar transformation: https://github.com/DataBiosphere/clinvar-ingest/blob/master/orchestration/templates/run-dataflow.yaml

orchestration/templates/run-extraction-pipeline.yaml
```diff
@@ -0,0 +1,38 @@
...
+          {{- with .autoscaling }}
+          - --autoscalingAlgorithm=THROUGHPUT_BASED
+          - --numWorkers={{ .minWorkers }}
+          - --maxNumWorkers={{ .maxWorkers }}
+          {{- end }}
+          {{- if .useFlexRS }}
+          - --flexRSGoal=COST_OPTIMIZED
+          {{- else }}
+          - --experiments=shuffle_mode=service
+          {{- end }}
```
snwessel, line 37: The args here are copied from the clinvar transformation: https://github.com/DataBiosphere/clinvar-ingest/blob/master/orchestration/templates/run-dataflow.yaml

orchestration/templates/cron-workflow.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - name: main
+        steps:
+          - - name: run-main-template
+              templateRef:
+                name: ingest-encode-data
+                template: main
+              arguments:
+                parameters:
+                  - name: gcs-prefix
+                    value: '{{ template "argo.timestamp" }}'
```
snwessel, line 26: I based this off of https://github.com/DataBiosphere/clinvar-ingest/blob/master/orchestration/templates/cron-workflow.yaml

orchestration/templates/run-extraction-pipeline.yaml
```diff
@@ -0,0 +1,38 @@
...
+          - name: gcs-bucket
+          - name: output-prefix
+      container:
+        image: us.gcr.io/broad-dsp-gcr-public/encode-extraction:{{ $version }}
+        command: []
+        args:
+          - --runner=dataflow
+          {{- $bucket := "{{inputs.parameters.gcs-bucket}}" }}
+             - --inputPrefix=gs://{{ $bucket }}/{{ "{{inputs.parameters.input-prefix}}" }}
+          - --outputPrefix=gs://{{ $bucket }}/{{ "{{inputs.parameters.output-prefix}}" }}
```
danxmoran, line 21: Double-check that this is the actual arg expected by the pipeline. I remember that the two are inconsistent; one uses "prefix" and the other uses "path", I think.

orchestration/values.schema.json
```diff
@@ -0,0 +1,70 @@
...
+    },
+    "serviceAccount": {
+      "type": "object",
+      "properties": {
+        "k8sName": { "type": "string" },
+        "googleName": { "type": "string" }
+      },
+      "required": ["k8sName", "googleName"]
+    },
+    "volumes": {
```
danxmoran, line 20: Don't need this `volumes` sub-schema

orchestration/values.yaml
```diff
@@ -0,0 +1,17 @@
+volumes:
+  storageClass: standard
+  downloadSize: 1Gi
+  extractSize: 10Gi
```
danxmoran, line 4: Don't need the default values for `volumes` either

orchestration/values.yaml
danxmoran: We can go more lightweight for ENCODE, I think. Let's start with `n1-standard-1` and see if we need to bump things up.

orchestration/templates/cron-workflow.yaml
danxmoran: Ah, wait: `argo.timestamp` won't be defined in this chart. The easiest thing to do would be to copy-paste the `_helpers.tpl` from clinvar. It'll eventually get bundled into a common chart that both can depend on...

orchestration/templates/_helpers.tpl
```diff
@@ -0,0 +1,27 @@
...
+
+{{/* Common retry settings to apply to flaky steps. */}}
+{{- define "argo.retry" }}
+retryStrategy:
+  retryPolicy: Always
+  backoff:
+    duration: 1s
+    factor: 2
+    maxDuration: 5m
+{{- end -}}
```
danxmoran, line 27: As you suggested in https://github.com/DataBiosphere/clinvar-ingest/pull/37#pullrequestreview-397554289, feel free to delete this file in favor of depending on the new common-templates chart.

PR #51, opened 2020-04-30
DSPDC-238 Add a "smoke test" mode to the orchestration pipeline

danxmoran:
Tested on a run that succeeded, but the plumbing is the same for the failure and error cases so I think it's ready for review.

### DIFF
--- a/orchestration/templates/cron-workflow.yaml
+++ b/orchestration/templates/cron-workflow.yaml
@@ -24,3 +24,31 @@ spec:
                 parameters:
                   - name: gcs-prefix
                     value: '{{ template "argo.timestamp" }}'
+      {{- if .Values.smokeTest.enable }}
+      - name: notify-if-failed
+        steps:
+          - - name: notify
+              template: send-slack-notification
+              when: '{{ "{{workflow.status}} != Succeeded" }}'
+      - name: send-slack-notification
+        container:
+          image: curlimages/curl:7.70.0
+          env:
+            - name: WORKFLOW_ID
+              value: '{{ "{{workflow.name}}" }}'
+            - name: WORKFLOW_STATE
+              value: '{{ "{{workflow.status}}" }}'
+            - name: SLACK_URL
+              valueFrom:
+                secretKeyRef:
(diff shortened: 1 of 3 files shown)


orchestration/templates/cron-workflow.yaml
```diff
@@ -24,3 +27,26 @@ spec:
...
+                  name: {{ .Values.smokeTest.slackUrl.secretName }}
+                  key: {{ .Values.smokeTest.slackUrl.secretKey }}
+          command: [curl]
+          args:
+            - -XPOST
+            - -H
+            - 'Content-type: application/json'
+            - -d
+            - '{"text": "Workflow $(WORKFLOW_ID) entered state: $(WORKFLOW_STATE)"}'
+            - $(SLACK_URL)
```
danxmoran, line 51: Slack's newest webhooks mechanisms tie the target channel to the URL, so the URL is all we need here.

orchestration/templates/cron-workflow.yaml
```diff
@@ -24,3 +27,26 @@ spec:
...
+                secretKeyRef:
+                  name: {{ .Values.smokeTest.slackUrl.secretName }}
+                  key: {{ .Values.smokeTest.slackUrl.secretKey }}
+          command: [curl]
+          args:
+            - -XPOST
+            - -H
+            - 'Content-type: application/json'
+            - -d
+            - '{"text": "Workflow $(WORKFLOW_ID) entered state: $(WORKFLOW_STATE)"}'
```
danxmoran, line 50: We can do all sorts of fancy things w/ message formatting here, but I wasn't sure what would be useful. Any ideas?

danxmoran:
See #monster-ci in Slack for the test message.

orchestration/templates/cron-workflow.yaml
snwessel: The `WORKFLOW_ID` would be something like `encode-ingest` right? If so, what you have makes sense to me.

quazi-broad: Hmm I'm wondering what other data points at the workflow level besides name and status?

PR #53, opened 2020-05-11
Sw dspdc 1045 diff tabular data

snwessel:
[Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1045)

Adding the diff-table step to encode (which should only run when a smoke test is not being run).

These changes are pretty heavily based off of the [clinvar equivalent](https://github.com/DataBiosphere/clinvar-ingest/blob/master/orchestration/templates/process-table.yaml) of this.

### DIFF
--- a/orchestration/Chart.yaml
+++ b/orchestration/Chart.yaml
@@ -7,6 +7,6 @@ type: application
 version: 0.0.0
 dependencies:
   - name: argo-templates
-    version: 0.0.4
+    version: 0.0.7
     repository: https://broadinstitute.github.io/monster-helm
     alias: argoTemplates
--- a/orchestration/Chart.yaml
+++ b/orchestration/Chart.yaml
@@ -7,6 +7,6 @@ type: application
 version: 0.0.0
 dependencies:
   - name: argo-templates
-    version: 0.0.7
+    version: 0.0.13
     repository: https://broadinstitute.github.io/monster-helm
     alias: argoTemplates
--- a/orchestration/templates/ingest-encode-data.yaml
+++ b/orchestration/templates/ingest-encode-data.yaml
@@ -51,16 +51,16 @@ spec:
                 - name: gcs-prefix
                   value: '{{ "{{inputs.parameters.gcs-prefix}}" }}'
                 - name: staging-bq-project
-                  value: '{{ .Values.bigquery.stagingData.project }}' # TODO define this value
+                  value: '{{ .Values.bigquery.stagingData.project }}'
                 - name: staging-bq-dataset
                   value: {{ $datasetName }}
                 {{- with .Values.bigquery.jadeData }}
                 - name: jade-bq-project
-                  value: '{{ .project }}' # TODO define this value
+                  value: '{{ .project }}'
                 - name: jade-bq-dataset
                   value: '{{ .dataset }}'
                 {{- end }}
-                {{- with .Values.repo }} # TODO define these values
+                {{- with .Values.repo }}
                 - name: url
                   value: '{{ .url }}'
                 - name: dataset-id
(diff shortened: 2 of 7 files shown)


orchestration/values.yaml
```diff
@@ -27,3 +27,6 @@ argoTemplates:
     name: gsutil-upload
 smokeTest:
   enable: false
+bigquery:
+  jadeData:
+    dataset: datarepo_broad_dsp_encode
```
snwessel, line 32: Does this dataset name look correct? I wasn't sure if it's already defined anywhere, so I just mirrored the name we we use for Clinvar

orchestration/templates/ingest-encode-data.yaml
```diff
@@ -35,3 +35,39 @@ spec:
...
+                  value: '{{ .Values.bigquery.stagingData.project }}'
+                - name: staging-bq-dataset
+                  value: {{ $datasetName }}
+                {{- with .Values.bigquery.jadeData }}
+                - name: jade-bq-project
+                  value: '{{ .project }}'
+                - name: jade-bq-dataset
+                  value: '{{ .dataset }}'
+                {{- end }}
+            withItems:
```
quazi-broad, line 63: This block is specifying the tables to loop over for this step of the DAG, right?
So this would run the process-table template once for every item as an input param for table-name?
Just wanted to confirm my understanding!

snwessel: Yeah, that's my basic understanding too. Each of the table names is passed in as a parameter using the `{{item}}` variable and I think there is an async task launched for each of them.

orchestration/values.yaml
danxmoran: It should match whatever value is used in the [build.sbt](https://github.com/DataBiosphere/encode-ingest/blob/master/build.sbt#L44)

orchestration/values.yaml
```diff
@@ -27,3 +27,6 @@ argoTemplates:
     name: gsutil-upload
```
danxmoran, line 27: You'll need to add an entry to this `argoTemplates` block to enable generation of the diff-tables `WorkflowTemplate`. See the latest clinvar-ingest.

orchestration/templates/process-table.yaml
```diff
@@ -0,0 +1,46 @@
...
+          - name: staging-bq-project
+          - name: staging-bq-dataset
+          - name: jade-bq-project
+          - name: jade-bq-dataset
+      dag:
+        tasks:
+          # 1) diff table
+          - name: diff-table
+            # WHEN the smoke test is not being run
+            when: '{{ not .Values.smokeTest.enable }}'
```
danxmoran, line 23: I would rather push the `when` to the other DAG, on `process-tables`. That way as more steps are added here we don't have to copy-paste this condition everywhere.

orchestration/Chart.yaml
```diff
@@ -7,6 +7,6 @@ type: application
 version: 0.0.0
 dependencies:
   - name: argo-templates
-    version: 0.0.4
+    version: 0.0.7
```
danxmoran, line 10: This is up to 0.0.13 now 😱

snwessel: Wow... I will bump it up. And no, I haven't run it yet, but I will give it a try now. It would just be the `apply-orchestration-workflow` script in monster-deploy, right?

danxmoran: Yes. You'll need to update the encode.yaml definition to pass in all the new parameters required by your branch.

snwessel: Looks like it worked :) I'm making another PR now for the parameters in encode.yaml in monster-deploy.

PR #56, opened 2020-05-26
DSPDC-1099 Add example input values for orchestration linting.

danxmoran:
Values are fake, but the structure / types are real. The new version of `MonsterHelmPlugin` will detect these files on `test` and validate that each one produces valid YAML with `helm template`.

### DIFF
--- a/orchestration/example-values/no-version.yaml
+++ b/orchestration/example-values/no-version.yaml
@@ -1,0 +1,25 @@
+gcs:
+  bucketName: bucket
+serviceAccount:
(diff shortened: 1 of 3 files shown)


danxmoran:
Yes, via `helm template` (it has built-in logic to validate that templated output is valid YAML). The behavior is defined [here](https://github.com/broadinstitute/monster-sbt-plugins/blob/master/plugins/helm/src/main/scala/org/broadinstitute/monster/sbt/Helm.scala#L70-L93).

PR #60, opened 2020-07-13
DSPDC-1160 Init README with high-level info.

danxmoran:
There's probably a lot more I could say here, but I'm not sure what would be most interesting...


danxmoran:
> Would a link to some document that summarizes the relationships between different entities be relevant to put here? Or at least a diagram?

Yes, I asked @kreinold to point me at any existing pictures of the ENCODE data model when she has the time. I forgot I asked her though, so thanks for the reminder!

PR #62, opened 2020-07-31
Sw dspdc 834 file ingest script

snwessel:
I tried to make this as organized as possible, but if you see room for improvement, let me know! Here's the [JIRA ticket](https://broadinstitute.atlassian.net/browse/DSPDC-834) for reference. 

A couple notes on this script:
- There is an optional `dataset_id` argument. If none is provided, the script will create a new dataset with a default name and schema, and then delete the dataset at the end. This is really helpful for testing!
- I had originally planned to have this script launch and monitor multiple ingest jobs asynchronously. Dan D suggested that it would be easier to just run the entire thing as one massive ingest job - so that's what it's doing now. 
-  I haven't done much testing to monitor time / cost yet because there needs to be a bit more work on the Jade side before we can do that. For now, I have just been testing it on a set of 10 files. 

I have it printing out some information as it runs, which looks something like this:
```
Creating a temporary dataset.
Waiting for job (ID: VIeF9QoASUqR6Cq09Y-ddw) to finish...
Job succeeded.
Created dataset with ID: 80802ff7-1333-4d19-ab4b-61958e2141e0

Submitting the bulk file ingest request
Waiting for job (ID: Jy71X_W4QfKgmUlc-JONqg) to finish...
Job succeeded.
Bulk file ingest finished in 40.893037948 seconds.
File result counts: 10 failed, 0 not tried, 0 succeeded

Deleting the temporary dataset.
Waiting for job (ID: MAqNZNTbQ9OSdDVTSIi3Kw) to finish...
Job succeeded.
Deleted dataset with ID: 80802ff7-1333-4d19-ab4b-61958e2141e0
```
The individual files are failing to ingest right now because a bug in the Jade API, but it looks like the script itself is working.


hack/bulk-ingest-explorer-files.py
```diff
@@ -0,0 +1,131 @@
...
+profile_id = sys.argv[1]
+is_production = (sys.argv[2] == 'prod') # default to dev
+control_file_path = sys.argv[3]
+dataset_id = sys.argv[4] if sys.argv[4:] else None # optional
+
+credentials, project = google.auth.default(scopes=['openid', 'email', 'profile'])
+authed_session = AuthorizedSession(credentials)
+Counts = namedtuple('Counts', ['succeeded', 'failed', 'not_tried'])
+jade_base_url = 'https://jade-terra.datarepo-prod.broadinstitute.org/' if is_production \
+    else 'https://jade.datarepo-dev.broadinstitute.org/'
```
snwessel, line 27: I was on the fence about the best way to take the environment (dev/prod) as an argument, and then get the URL to use based on that. Let me know if you have any suggestions for how this could be improved.

hack/bulk-ingest-explorer-files.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+# submit a bulk file ingest request. If successful, return the id of the job.
+def request_bulk_file_ingest(target_dataset_id: str, load_tag: str, profile_id: str, control_file_path: str):
+    print('\nSubmitting the bulk file ingest request')
+    args = {
+        'loadTag': load_tag,
+        'profileId': profile_id,
+        'loadControlFile': control_file_path,
+        'maxFailedFileLoads': 1000000  # the jade api will eventually have 'unlimited' as the default
```
snwessel, line 103: The api currently fails when you don't specify maxFailedFileLoads. Dan D suggested that the default value for this should eventually be fixed and be updated to be "unlimited". We want the value to at least be larger than the number of files we're ingesting (so it doesn't leave the job partially finished)... I have it set to 1,000,000 right now, even though that feels a little hacky.

hack/bulk-ingest-explorer-files.py
quazi-broad: I tried looking in other Python scripts in our projects and it looks like we use:
`base_url = os.environ["API_URL"]`
Was looking at this script:
https://github.com/DataBiosphere/clinvar-ingest/blob/master/orchestration/scripts/poll-ingest-job.py
Looks like you might be able to do the same with some of the other arguments, I'm just forgetting where we  would make the change to set that API_URL

hack/bulk-ingest-explorer-files.py
quazi-broad: Understood, so that means we're waiting on a change from Jade before we switch this ti "unlimited" or remove it completely, right?

hack/bulk-ingest-explorer-files.py
snwessel: Good point. I think for those, it's being set as an env variable when the scripts are called (like [here](https://github.com/DataBiosphere/clinvar-ingest/blob/b6e5b995fae6de5583b6ac722375a5a68eca9392/orchestration/templates/process-xml-release.yaml#L208)). 

But I agree that getting the url itself in as the argument (instead of "dev" or "prod") would make more sense.

hack/bulk-ingest-explorer-files.py
snwessel: I'm not sure when Jade is going to fix/update it on their end, and the script works with the hard coded value, so I would be inclined to leave it as it is. If they happen to update it sooner, that would be great, but I wouldn't want to put this on hold until they do that. 

Dan D had suggested using the value of 1,000,000 for now since that's the max number of files Jade allows to be ingested in a single job.

hack/bulk-ingest-explorer-files.py
snwessel: Just updated it! So now it looks something like this to run the script on the command line:
```python3 bulk-ingest-explorer-files.py <profile_id> <base_url> <control_file_path> <dataset_id>```

quazi-broad: nice, that looks god. so this script is just being called manually for the time being? oh actually I think I see some tickets to update the code to call this..

quazi-broad:
Looks good to me!

hack/bulk-ingest-explorer-files.py
snwessel: Yeah, I have just been running it manually. In theory, this ingest will only be run once, so we don't need to set it up with too much infrastructure. 

My understanding is that a bunch of the encode files were copied over to a Broad-owned Google bucket, when they were building the "Encode Explorer" (when monster had just become a team). The Encode Explorer isn't updated anymore so we would only need to copy the files out of it once. But, it's easier (and maybe cheaper?) to move the big set of files from the encode explorer to jade, than to move them from the Encode site itself to Jade (because GCS -> GCS is easier than AWS -> GCS). 

So the next ticket is to do some performance testing and then do the _single run_ of this script in production that will copy the ~50,000 files or (about 95 TB).

PR #63, opened 2020-12-14
Check for '90 or above' age identifier

aherbst-broad:
## Why
Our transformation pipeline is failing on the string "90 or above" in the age field.

## This PR
* Adds a check for this string
* Adds a basic test to make sure we can parse records of this form


ThePletch (approved):
data ingest may be the most sisyphean field in all of software development

raaidbroad (approved):
simple, stellar ⭐

PR #64, opened 2020-12-15
Pull in latest ingest-utils

aherbst-broad:
## Why
`ingest-utils` 2.1.6 pulls in latest scio; `encode-ingest` is far behind and should be brought up to date.

## This PR
* Upgrades to `ingest-utils` 2.1.6
* Fixes a breaking change when writing json lists

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -1,7 +1,0 @@
-coverage:
-  status:
-    project:
-      default:
-        target: 75%
-        threshold: 5%
-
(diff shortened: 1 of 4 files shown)


.github/codecov.yaml
```diff
@@ -1,7 +0,0 @@
-coverage:
```
aherbst-broad, line 1: We still get code coverage via their default settings, this just removes the status checks.

ThePletch (approved):

raaidbroad (approved):

PR #65, opened 2020-12-17
Always notify slack

aherbst-broad:
## Why
We're only pinging slack during our smoke tests; we should ping slack regardless of smoke testing or prod and indicate the environment in the message.

## This PR
* Reworks the slack notification to ping the slack API with a bearer token
* Pulls the token from a new secret in k8s (already in place)


raaidbroad (approved):

quazi-broad (approved):
Changes look pretty straightforward.

PR #70, opened 2021-03-24
[DSPDC-1670] Update to latest ingest-utils to pull in latest graal vm image

aherbst-broad:
## Why
We cannot publish an encode docker image because the base graal vm image has vanished.

## This PR
* Updates to the latest version of ingest-utils which pulls in an updated docker image.


eparkmonster (approved):

quazi-broad (approved):
lgtm!

PR #71, opened 2021-03-24
Make this field optional as it is missing in some of the current ingest rows

aherbst-broad:
## Why
The encode transformation pipeline is blowing up due to some rows missing an assay_silms value. 
[Relevant ticket](https://broadinstitute.atlassian.net/secure/RapidBoard.jspa?rapidView=495&projectKey=DSPDC&view=planning&selectedIssue=DSPDC-1671&issueLimit=100)

## This PR
* Converts the field to be not required and updates the transformation code
* I have run this against the current dataset in our dev gs bucket and verified that it runs to completion.

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 75%
+        threshold: 5%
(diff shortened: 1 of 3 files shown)


eparkmonster (approved):

quazi-broad (approved):
changes look straightforward

PR #80, opened 2022-07-08
encode updates for TIM model

ahaessly:


aherbst-broad (commented):
Thanks for doing this @ahaessly . +2651/-1247 is a pretty hefty diff -- would you be open to a working session to walk us through these changes?

aherbst-broad (approved):

